PR #1, opened 2020-02-28
Replace fetch with XMLHTTPRequest

stephanebisson:
While `fetch` is very convenient it is not supported by all browsers.

This change uses XMLHTTPRequest to query the page summary endpoint and gets rid of the `unfetch` library. It also uses callbacks instead of promises to remove the need for a promise polyfill in IE.

https://phabricator.wikimedia.org/T243774


jpita:
I think this broke IE11, there's no popup now

PR #4, opened 2020-06-03
Publish compiled output to gh-pages

stephanebisson:
https://phabricator.wikimedia.org/T240012

When there's a new commit on the master branch, this CI configuration will generate the demo page and publish it to the gh-pages branch.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,7 +34,8 @@ jobs:
       - run:
           name: Deploy to gh-pages
           command: |
-            gh-pages --dist ./demo
+            cd demo
+            gh-pages --dist .
 workflows:
   version: 2
   build:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,6 +18,7 @@ jobs:
     docker:
       - image: node:8.10.0
     steps:
+      - checkout
       - attach_workspace:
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -6,8 +6,48 @@ jobs:
...
+  update-demo:
+    docker:
+      - image: node:8.10.0
+    steps:
+      - checkout
+      - attach_workspace:
+          at: .
+      - add_ssh_keys:
+          fingerprints:
+              - "c3:03:93:79:5d:c7:23:b8:4a:4d:5d:41:e3:38:3f:84"
```
medied, line 32: Trying to understand more about this fingerprint..

https://superuser.com/questions/421997/what-is-a-ssh-key-fingerprint-and-how-is-it-generated
https://circleci.com/docs/2.0/configuration-reference/#filters

So it's meant to stay public on master?

hueitan: ya I also have the same problem and who owns this fingerprint?

stephanebisson: There's a key that allows circleci to push to github. The private part is own by circleci and the public part is on github. The fingerprint is just to identify which key to use. This is not sensitive.

PR #5, opened 2020-06-06
T251265: update preview discovery

medied:
Phab ticket: https://phabricator.wikimedia.org/T251265

Goal is well explained in the phab ticket. There’s some CSS details that we need to decide on with Sudhanshu because browsers differ in rendering behaviors, for example:

* I'm leaving `vertical-align` as `top` to start because it looked more centered after some visual tests [1]
* `padding` sometimes looks better with one set of values in one browser and then different in another [2]
* Should we be setting `line-height`? Depending on screen size and padding sometimes I see stuff like this [3]

---

[1] 
image

[2]
image

[3]
image


style/index.less
```diff
@@ -0,0 +1,18 @@
+.content {
```
stephanebisson, line 1: Note that `.content` is relevant to our demo page, not so much for external consumer of the library. Same for the margin below.

style/index.less
```diff
@@ -0,0 +1,18 @@
...
+  margin: 20px 10%;
+
+  [data-wikipedia-preview] {
```
stephanebisson, line 4: According to the documentation, the [selector is configurable](https://github.com/wikimedia/wikipedia-previews#options-of-the-init-function). 

We could attach this style to another class, then apply this class in the init function when we discover all the element that should have a popup attached to them.

style/index.less
```diff
@@ -0,0 +1,18 @@
...
+    background-color: #eaf3ff;
+    padding: 2px 2px 2px 2px;
```
stephanebisson, line 6: This padding (and the :after icon) will make the whole text jump since this is applied after the initial page rendering.

style/index.less
```diff
@@ -0,0 +1,18 @@
...
+    font-family: 'Linux Libertine','Georgia','Times',serif;
+    border-radius: 4px;
+    border: solid 1px rgba(51, 102, 204, 0.48);
+
+  }
+
+  [data-wikipedia-preview]:after {
+    content: url(/images/info_24_px.svg);
```
stephanebisson, line 14: This URL makes sense in the context of the dev server but this is a standalone library. At runtime, this URL will try to be resolved relative to the partners website.

style/style.less
```diff
@@ -0,0 +1,2 @@
+@import './popup';
```
stephanebisson, line 1: I like this change but `index` should probably be the one that includes the others.

style/index.less
medied: Yea the padding and icon are some of the CSS specifics that Sudhanshu and I are still refining. What do you mean by "make the whole text jump" exactly?

style/index.less
medied: You’re right, we could feed it a relative path then

stephanebisson: Remember that this library is distributed as a single js file. This icon needs to be embedded in it.

style/index.less
stephanebisson: The content of the partner site will be first rendered, then this library will be loaded at the end of the body and this style will be applied. When it does, all those links will become 4px wider, which will make most of the text move. 

Similar to when you have an image that loads late without an appropriate placeholder: everything below the image jump down.

style/index.less
medied: Right, leaving the `.content` css in index.html then

style/index.less
medied: Good call, I'm applying it now with [pure javascript](https://stackoverflow.com/questions/6787383/how-to-add-remove-a-class-in-javascript) since `classList` is [not supported in IE](https://developer.mozilla.org/en-US/docs/Web/API/Element/classList#Browser_compatibility)

style/index.less
medied: I see, I just tested doing this and it's working for me in Firefox, Chrome, Safari. Would this be an appropriate solution though? Haven't committed it yet. 

```
content: url("data:image/svg+xml;utf8,<svg xmlns='http://www.w3.org/2000/svg' width='20' height='20' viewBox='0 0 20 20'><g fill='none' fill-rule='evenodd'><path d='M0 0L20 0 20 20 0 20z'/><path fill='%2336C' fill-opacity='.7' fill-rule='nonzero' d='M10.556 1.111c-4.6 0-8.334 3.733-8.334 8.333s3.734 8.334 8.334 8.334 8.333-3.734 8.333-8.334-3.733-8.333-8.333-8.333zm.833 12.5H9.722v-5h1.667v5zm0-6.667H9.722V5.278h1.667v1.666z'/></g></svg>");
```

https://stackoverflow.com/questions/10768451/inline-svg-in-css

stephanebisson: I think so but webpack or less should be able to do it for us.

https://github.com/atlassian/less-plugin-inline-svg

medied: Agree using a plugin would be a cleaner approach. I'm following [this documentation](https://webpack.js.org/loaders/less-loader/#plugins) to enable it but I'm not getting any results or errors. It looks like [there have been issues](https://github.com/webpack-contrib/less-loader/issues/188) with `less-loader` and plugins before that might be affecting here, unless I'm missing something. Any ideas? I can file an issue with `less-loader` or `less-plugin-inline-svg` folks to try our luck there.

stephanebisson: Yeah, let it go if it doesn't work. We'll figure it out later if we have more image and it becomes an annoyance to manage them inline.

stephanebisson:
> Couple of comments
> 
> 1. Should we add hover pointer to the style? currently it is cursor

Good point(er). You should check with design on phab.

> 2. `popup` `style` css have two different prefix `wp-*` `wikipedia-preview`, it would be great if we can use the same prefix to prevent conflict and easier to maintain

We have to think hard about all the classes to make sure they are 1) not too long but 2) very unlikely to conflict with partners sites. Try to keep it consistent in this PR but don't worry too much about the prefix for now.

> 3. style can't be changed with `data-wikipedia-preview`, we should remove it from the README section
>    (we can wait until we have more content to do the change for README file, furthermore, if this pr gets merge to master, we should mention the library is under development)

This style needs to be configurable (enable/disable) by partners. The documentation needs to stay up to date in each PR.

All code is generally under construction. We're working towards a new version. The previous one will continue to exist on npm.

style/index.less
medied: Gotcha. I'm researching more about what we discussed earlier this week, see if it's possible to load the style in the header first and then the script at the end of the body. 

From what I can gather after some initial investigation, according to this `style-loader` docs [1] the default behavior is that it inserts the CSS in the `<head>` tag already. Inspecting the HTML source locally I do see `style` tags with the CSS, but I need to look into this further to actually confirm/verify the load order. 

[1] - https://github.com/webpack-contrib/style-loader#insert

medied:
Good points all around, some thoughts at the moment:

* Checked in with design on Phab: 
    * https://phabricator.wikimedia.org/T251265#6232700
* Agree to try to keep class names consistent
* Regarding how to make the styles configurable (enable/disable):
    * I still need to understand a bit better how exactly the library will be used by partner but what I'm going to explore further is: 
        * can we do something with `argv` similar to how we’re using `mode`
        * where we read a specific flag to either enable or disable styles (where if disabled, that means the user would be personalizing the design on their own)
        * making use of `exclude` somehow
            * https://stackoverflow.com/questions/44376589/webpack-exclude-a-specific-file
            * https://webpack.js.org/configuration/module/#ruleexclude

Will report back where that takes me, but welcome any early feedback/reactions

stephanebisson:
> * Regarding how to make the styles configurable (enable/disable):
>   
>   * I still need to understand a bit better how exactly the library will be used by partner but what I'm going to explore further is:
>     
>     * can we do something with `argv` similar to how we’re using `mode`
>     * where we read a specific flag to either enable or disable styles (where if disabled, that means the user would be personalizing the design on their own)
>     * making use of `exclude` somehow
>       
>       * https://stackoverflow.com/questions/44376589/webpack-exclude-a-specific-file
>       * https://webpack.js.org/configuration/module/#ruleexclude

Webpack can be misleading here. The component is compiled with webpack _once_ and then published to npm for everyone to use. I see 2 ways for partners to have some control over the style

1. Include, or not, a css file provided by us that includes this style
2. A new config option in the `wikipediaPreview.init()` function but that will execute late and may cause the text to jump.

medied:
By the way I'm thinking about enabling popoup on tap/click from mobile in this PR to work on [this request](https://phabricator.wikimedia.org/T251265#6234909) from design:

>  ...Wikipedia Preview's hyper link state once user has interacted with it. This won't require any additional design just the color of a link will change after tap/click on it.

style/style.less
```diff
@@ -0,0 +1,14 @@
+.wp-style {
+  background-color: #eaf3ff;
+  padding: 2px 2px 2px 2px;
+  font-family: 'Linux Libertine','Georgia','Times',serif;
+  border-radius: 4px;
+  border: solid 1px rgba(51, 102, 204, 0.48);
+  cursor: pointer;
+
+  &:after {
+    content: url("data:image/svg+xml;utf8,<svg xmlns='http://www.w3.org/2000/svg' width='20' height='20' viewBox='0 0 20 20'><g fill='none' fill-rule='evenodd'><path d='M0 0L20 0 20 20 0 20z'/><path fill='%2336C' fill-opacity='.7' fill-rule='nonzero' d='M10.556 1.111c-4.6 0-8.334 3.733-8.334 8.333s3.734 8.334 8.334 8.334 8.333-3.734 8.333-8.334-3.733-8.333-8.333-8.333zm.833 12.5H9.722v-5h1.667v5zm0-6.667H9.722V5.278h1.667v1.666z'/></g></svg>");
```
hueitan, line 10: is there any loader so we can import the svg image here, then loader convert it to this format?

_I also need this loader in the layout ticket_

medied: Check out [this thread](https://github.com/wikimedia/wikipedia-preview/pull/5#discussion_r439390548), I tried with this plugin but ran into issues: https://github.com/atlassian/less-plugin-inline-svg

hueitan: Thanks! it works for me, checkout this commit
https://github.com/wikimedia/wikipedia-preview/pull/12/commits/64ee96d410b48abbd074259355f7028f2c5bc695

medied:
> 1. Include, or not, a css file provided by us that includes this style

https://github.com/wikimedia/wikipedia-preview/pull/5/commits/037ce42ff779c29dbec947ba51383930e3400903 gets us the following expected behavior:

* Style for the links distributed as a distinct css file (`wikipediaPreview.css`)
  * This file can be included in the header if default style is desired, or removed from header if personalized style is desired
* Style for the popup and preview bundled with the js so it's always included.
  * I was able to differentiate between link styles and popup styles by making the link styles have its [own entry point](https://stackoverflow.com/questions/55677070/webpack-mini-css-extract-plugin-output-multiple-css-files-on-single-entry) as a .css file (instead of .less)
* No text jumping
* Less generic class name

Note the same behavior is achieved if we attach the link style to the selector `[data-wikipedia-preview]`, in which case the class name does not need to be added, but then this conflicts with the selector name being customizable (as noted in README) so I’m attaching it to a class name for now.

medied:
@hueitan @stephanebisson ready for review

README.md
```diff
@@ -21,7 +21,7 @@ It is a small Javascript component that does not have any dependencies and shoul
 ### Standalone Script
 
 ```html
-<script src="wikipedia-preview.production.js"></script>
+<script src="wikipedia-preview.index.production.js"></script>
```
stephanebisson, line 24: It would be ideal if you could avoid renaming this file.

README.md
```diff
@@ -72,23 +72,37 @@ wikipediaPreview.init({
...
+
+To use a language different than the language specified in the options, use the `data-wp-lang` attribute.
+
+#### .wmf-wp-with-preview
+
+To use the default trigger link styling:
+
+* Add the following link to the page header:
+```html
+<link href="dist/wikipediaPreview.css" rel="stylesheet">
```
stephanebisson, line 89: The official doc shouldn't mention the `dist/` folder.

style/style.css
```diff
@@ -0,0 +1,14 @@
+.wmf-wp-with-preview {
```
stephanebisson, line 1: This file should probably be renamed. I don't have a good suggestion at the moment but style.css doesn't say anything about what this is. Same with css.js.

medied:
> It would be ideal if you could avoid renaming this file.
> ...
> This file should probably be renamed

Okay, how about 0b4a429c5b494a5384c81a265671bd543de6ef1f

README.md
medied: One alternative is simply not including the HTML example and updating text to something like:

> Add a `<link>` in your page header referencing the `wikipediaPreview.css` asset

Just curious, why shouldn't the doc mention the `/dist` folder?

stephanebisson: I think the HTML example is fine. 

The dist/ folder makes sense for us in the source repo but when we distribute, only the compiled output should be provided.

medied: Hmm right, how about just removing `dist/` then and have it be consistent with what's under 'Standalone Script' section?

```html
<link href="wikipediaPreview.css" rel="stylesheet">
```

PR #6, opened 2020-06-14
webdriverio with crossbrowsertesting

jpita:
This branch has one [selenium](https://github.com/SeleniumHQ/selenium) test running with [webdriverio](https://webdriver.io) against [crossbrowsertesting](https://crossbrowsertesting.com) browsers.

This allows us to run tests against a [multitude](https://crossbrowsertesting.com/api/v3/selenium/browsers?format=json) of browsers.

If you have a login for the app you can see the results and video recordings of the tests [here](https://app.crossbrowsertesting.com/selenium/results).

Right now we have tests running for:
* Chrome on Windows
* Firefox on Windows
* Safari on Mac
* Edge on Windows
* Internet Explorer 11 on Windows

In order to add more we just need to create more wdio.BROWSER.conf.js files with the browser capabilities.

The test does the following actions:
* Go to the demo app
* Hover over the link
* Click on the image
* Click on the link
* Check that the new tab opens on the right article

Feel free to submit improvements to the code, it's my first time using both webdriverio and crossbrowsertesting and as we all know my knowledge (and love) of JS is very limited.


.circleci/config.yml
```diff
@@ -6,8 +6,63 @@ jobs:
...
-            npm install
-            npm run build
-            npm run coverage
+            npm install -g --silent gh-pages
+            git config user.email "ci-build@wikimedia.org"
+            git config user.name "ci-build"
+      - run:
+          name: Deploy demo
+          command: |
+            gh-pages --dist demo
```
jpita, line 57: @stephanebisson is this deploying the demo app?
I see this is not on master, I think that's why I branched from #4 .

.circleci/config.yml
```diff
@@ -6,8 +6,63 @@ jobs:
...
+  version: 2
+  build:
+    jobs:
+      - build
+      - update-demo:
+          requires:
+            - build
+          filters:
+            branches:
+              only: master
```
jpita, line 68: if this deploys only on master maybe we can create a new demo page just for tests that deploys on every build?

stephanebisson: Ideally the tests should run against the local version of the code

jpita: The local connection is not reliable enough for us to trust it IMO.
Deploying the code on a test page should be fairly simple, no?

jpita: can we have a test environment to run the tests against?

hueitan: we use github gh-pages serve as the master environment for the demo purpose, unfortunately it doesn't support extra environment. We would need to use local version for the CI test.

hueitan:
can we move all the `wdio.*.conf.js` files under `test/config/` folder, that easier for us to add more test conf in the future and don't get longer in the root directory.

.circleci/config.yml
jpita: 😢

stephanebisson: All PRs need to be tested independently so we would need a distinct environment for each. And we need to be able to clean them up at some point. It's possible but I'm not sure what a simple solution could be.

Testing against local is so much better and simpler. CircleCI has a large collection of images with different browsers. Maybe worth looking if we can find some IEs in there.

Also, when you say the local connection is not reliable, what do you mean, what do you see? Maybe there's something we can do to help it.

jpita: I've opened a support ticket with cbt, it's a problem on their side (they open a tunnel to where the code is running and it doesn't seem to work very well/slow).

If we can run the tests inside circleci would be perfect, I just need to know the exact version of the browser that is running in the image.

jpita:
> can we move all the `wdio.*.conf.js` files under `test/config/` folder, that easier for us to add more test conf in the future and don't get longer in the root directory.

done

jpita:
can you review again?
added more code to set the pass/fail in cbt

test/specs/selenium.js
```diff
@@ -18,4 +20,21 @@ describe('wikipedia preview - ', () => {
...
+    afterEach(async function(){
+        var score=this.currentTest.state.split('ed')[0]
+        try {
+            await axios.put(
+                'https://crossbrowsertesting.com/api/v3/selenium/' + browser.sessionId,
+                {'action': 'set_score', 'score': score },
+                {
+                    auth: {username: browser.config.user, 
+                           password: browser.config.key
+                        }
```
hueitan, line 33: Can you try this? if it works then we don't need to use `axios` dependency

```es6
await fetch('https://crossbrowsertesting.com/api/v3/selenium/' + browser.sessionId, {
    method: 'PUT', 
    headers: {
        'Authorization': 'Basic ' + btoa(`${browser.config.user}:${browser.config.key}`)
    },
    body: JSON.stringify({'action': 'set_score', 'score': score })
})
```

jpita:
current status of the build:

- chrome passes
- edge passes
- safari fails due to cbt tunnel not working
- firefox fails due to cbt tunnel not working 
- IE fails due to a bug in the code

jpita:
@hueitan the unit tests are failing after the merge, can you check what I did wrong?

jpita:
can anyone fix the conflicts for this please?
make sure it is correct because I did a merge before and fixed the conflicts myself.

hueitan:
> can anyone fix the conflicts for this please?
> make sure it is correct because I did a merge before and fixed the conflicts myself.

things completely wrong even after I merge the latest master, why is that?

stephanebisson:
Won't be pursued in the short term.

PR #9, opened 2020-06-18
T254212 Fix mouse leave event

hueitan:
Phabricator ticket : https://phabricator.wikimedia.org/T254212

Problem
===

The popup close immediately as soon as the mouse leaving the hover target element

Solution
===

FF doesn't have `event.toElement`, use `event.relatedTarget` instead

Note
===

it works on Firefox 77 / Firefox 58 (older)


hueitan:
Update and ready to be reviewed in FF.

PR #12, opened 2020-06-23
New Layout

hueitan:
Phabricator Link : https://phabricator.wikimedia.org/T250796

New layout for Wikipedia Preview


hueitan:
Status of latest commit : add `isMobileDevice` constant, and support different layout/event when dealing mobile or desktop device.
Next commit : Layout CSS for mobile and desktop less

stephanebisson:
**Please note that mobile mode is NOT in scope for this task.**

The goal here is to finish the new layout to unblock mobile positioning, gallery view, touch interactions, RTL, TWN, etc. Everything else depends on it.

hueitan:
Yes, Mobile is not fully supported yet in this PR, this toggle allows me thinking more about the structure and how it looks like in the beginning with very minimal support.

---

The overall layout code is ready although it looks ugly now, we may need to improve when working on the development script / newer task / feature support. The styling also not 100% supported yet, will push more changes here for the final, this is ready for review and some `@todo` line within the code also needed to update either in this or later pr.

| Desktop (Hover) | Desktop (Hover w/ few content) | Mobile (Hover)
| --- | --- | --- |
| desktop-hover | desktop-hover-less-content | mobile-hover |

hueitan:
> individual components remain small and unit-testable

the unit tests fail due to the mix of popup/preview elements/event bindings, will think of the solution to make the components stay individual

stephanebisson:
> > individual components remain small and unit-testable
> 
> the unit tests fail due to the mix of popup/preview elements/event bindings, will think of the solution to make the components stay individual

It's important that the components remain relatively decoupled and that the unit tests remain robust with high coverage.

"Preview" is the actual preview of an article. We should be able to feed it article data and unit test that is shows it correctly and it has the correct state (collapsed, expanded) based on the size of the data and that it is able to transition between those states. It should not know about the API layer.

"Popup" is the box that shows the preview but really it doesn't know about it. It could show anything. It is responsible proper positioning. Current unit tests on this are not too bad, I think. When we do mobile (NOT IN SCOPE HERE) we can update "Popup" to handle it or, even better, create another container that acts as a bottom drawer and let the index decide which container to use depending on the platform.

hueitan:
Add the `customEvent` and `subscribe` events between popup container and preview elements

using event this way may bring some side effect, but it's a good start to have popup, preview, bridge (here is event) as the seperate

Next : add new unit test and update layout css

hueitan:
I think it's ready for review

src/index.js
```diff
@@ -1,17 +1,20 @@
...
-	const globalLang = lang || 'en'
-	root = root || document
-	selector = selector || '[data-wikipedia-preview]'
-	popupContainer = popupContainer || document.body
-
+function init({
+	root = document, 
+	selector = '[data-wikipedia-preview]', 
+	lang = 'en', 
+	popupContainer = document.body }) {
```
medied, line 11: Are we leaving the `}) {` part at the end of this line intentionally?

style/popup.less
```diff
@@ -1,71 +1,156 @@
...
 	z-index: 110;
+	overflow: hidden;
+	position: absolute;
 
-	* {
-		box-sizing: border-box;
-	}
-
-	.wp-article-preview {
+	.wikipediapreviews {
```
medied, line 11: Do we plan to separate the popup CSS from the preview CSS in the future?

src/index.js
```diff
@@ -1,17 +1,20 @@
...
+	const globalLang = lang
 	const popup = createPopup(popupContainer)
+	const events = customEvents(popup)
+	popup.subscribe(events)
```
medied, line 15: Could someone please give me a high level of the relationship between `popup` and `events`, the reasoning behind subscribing to events?

medied:
Well done to both you, this is looking really nice

src/index.js
hueitan: > "Preview" is the actual preview of an article. We should be able to feed it article data and unit test that is shows it correctly and it has the correct state (collapsed, expanded) based on the size of the data and that it is able to transition between those states. It should not know about the API layer.

> "Popup" is the box that shows the preview but really it doesn't know about it. It could show anything. It is responsible proper positioning. Current unit tests on this are not too bad, I think. When we do mobile (NOT IN SCOPE HERE) we can update "Popup" to handle it or, even better, create another container that acts as a bottom drawer and let the index decide which container to use depending on the platform.

"Event" is the central place where the listener binds and removes, it uses as the event connection between Popup and Preview. For example, the event subscription tells popup to close when the preview close button is clicked.

style/popup.less
hueitan: ya, it's good to have those separately. I will do it in this PR.

hueitan: one thought of it! current CSS structure we have is `index` that includes `popup` (popup includes preview style)

my suggestion change:
no more `index.less` needed as `popup.less` (import in `popup.js`) `preview.less` (import in `preview.less`)

what do you think?

stephanebisson: Makes sense to me if it works well at runtime

medied: Sounds good to me

style/preview.less
```diff
@@ -0,0 +1,145 @@
...
+            &-readmore {
+                display: none;
+            }
+
+            &-readonwiki {
+                display: block;
+            }
+        }
+    }
+}
```
stephanebisson, line 145: missing new line

PR #13, opened 2020-06-25
selenium tests running in parallel only on master against demo page

jpita:
selenium tests running in parallel only on master against demo page


.circleci/config.yml
```diff
@@ -39,7 +39,18 @@ jobs:
       - run:
           name: Deploy demo
           command: |
-            gh-pages --dist demo -m "Update demo [ci skip]"
```
jpita, line 42: @stephanebisson not sure why this line is changed...
is this correct?

stephanebisson: We need the original version with the custom message.

jpita: fixed

test/unit/api.test.js
```diff
@@ -1,5 +1,5 @@
 const assert = require('assert')
-const { requestPagePreview } = require('../src/api')
+const { requestPagePreview } = require('../../src/api')
```
jpita, line 2: @stephanebisson same with this, not sure why is changed....

jpita: nevermind, this is due to the changed folders in the test folder

test/e2e/specs/selenium.js
```diff
@@ -0,0 +1,39 @@
...
+
+describe('wikipedia preview - ', () => {
+
+    it('should open article on another tab', async (done) => {
+        await browser.maximizeWindow()
+        await browser.url('https://wikimedia.github.io/wikipedia-preview')
+        el = await $('.content-en > p:nth-child(2) > span:nth-child(2)')
+        await el.moveTo()
+        await el.click()
+        el=await $('div.wp-text-content > div.wp-title')
```
stephanebisson, line 12: You should have spaces around = like on line 9. Also applies to lines 15 and 17.

test/e2e/specs/selenium.js
```diff
@@ -0,0 +1,39 @@
+const assert = require('assert')
+const axios = require('axios');
+
+describe('wikipedia preview - ', () => {
+
+    it('should open article on another tab', async (done) => {
+        await browser.maximizeWindow()
+        await browser.url('https://wikimedia.github.io/wikipedia-preview')
+        el = await $('.content-en > p:nth-child(2) > span:nth-child(2)')
```
stephanebisson, line 9: This is very low-level. Did you consider a page object or a private function to encapsulate what this represents?

test/e2e/config/wdio.conf.js
```diff
@@ -0,0 +1,52 @@
...
+        record_network: 'false'
+    },{
+        name: 'wikipedia preview safari on: '+process.env.ENVIRONMENT,
+        build: '1.0',
+        platform: "Mac OSX 10.15",          // Gets latest version by default
+        browserName: 'safari',     // To specify version, add version: "desired version"
+        record_video: 'true',
+        record_network: 'false'
+    }]
+}
```
stephanebisson, line 52: There should be a new line a the end of this file.

test/e2e/specs/selenium.js
jpita: how can I use the linter for this?

test/e2e/specs/selenium.js
jpita: a page object is only worth it when we use the same selector more than once.
since right now we only have one test I think is not needed.

test/e2e/config/wdio.conf.js
jpita: same as before, how can I add a linter for this?

test/e2e/specs/selenium.js
hueitan: _we need the linter! please wait for few days until we include it in the master._

jpita: fixed

test/e2e/specs/selenium.js
stephanebisson: There's value in knowing what this selector is supposed to target. It makes the test... readable.

stephanebisson: This is the selector you need: `'.content-en [data-wikipedia-preview]:nth-child(2)'`

It targets the second article with a preview in the english content text.

package.json
```diff
@@ -4,26 +4,35 @@
...
+    "@wdio/spec-reporter": "5.23.0",
+    "axios": "^0.19.2",
     "babel-loader": "^8.0.6",
     "bufferutil": "^4.0.1",
     "canvas": "^2.6.1",
     "css-loader": "^3.4.2",
     "eslint": "^6.8.0",
     "eslint-loader": "^3.0.3",
+    "expect": "^26.0.1",
+    "expect-webdriverio": "^1.1.5",
```
hueitan, line 35: never tried but asking here, do we use `expect` `expect-webdriverio` `@wdio/mocha-framework` here? or does it require in the current status?

PR #15, opened 2020-06-30
Update screenshot of the new layout

hueitan:
as title


medied:
Should we leave this image outside of the `/images` folder?

stephanebisson:
> Should we leave this image outside of the `/images` folder?

It should not be mixed with the images used by Preview itself.

PR #16, opened 2020-06-30
add linter for script and mocha test file

hueitan:
use [eslint-config-wikimedia](https://github.com/wikimedia/eslint-config-wikimedia) as the linter


stephanebisson:
I would like to propose that we disable the rule that forces semi-colons. Unlike MediaWiki, our code is transpiled and semi-colons are added automatically.

hueitan:
> I would like to propose that we disable the rule that forces semi-colons. Unlike MediaWiki, our code is transpiled and semi-colons are added automatically.

True, we don't need `;` in our code. rule added.

Regards the mocha test rule, I don't modify it as we follow the config from wikimedia, what do you say?

package.json
```diff
@@ -6,6 +6,7 @@
   "scripts": {
     "test": "mocha --require ignore-styles,@babel/register",
     "test:watch": "mocha --require ignore-styles,@babel/register -w",
+    "test:fix-lint": "eslint test --fix",
```
stephanebisson, line 9: I don't think this task should be called "test"... maybe just "lint" or "eslint"

hueitan: This is the linter for the folder test, and it has it’s owned eslint config, do you think we also need to apply no semi in the test spec?

Note that client code and test code use different eslint config

stephanebisson: Oh, I didn't notice that. Why do they have different configs?

hueitan: There is config for client code, 

and one for mocha, i think that’s the reason mocha has the variable describe and it, and runs with node script

https://github.com/wikimedia/eslint-config-wikimedia

stephanebisson:
> Regards the mocha test rule, I don't modify it as we follow the config from wikimedia, what do you say?

I don't like having semis in one part of the code and not in the other. I would prefer to omit them everywhere.

hueitan:
> > Regards the mocha test rule, I don't modify it as we follow the config from wikimedia, what do you say?
> 
> I don't like having semis in one part of the code and not in the other. I would prefer to omit them everywhere.

Done 💪

PR #17, opened 2020-06-30
add stylelint-config-wikimedia for style linter

hueitan:
_no more bad format from me after the linter 🔢_


.stylelintrc.json
```diff
@@ -0,0 +1,3 @@
+{
+	"extends": "stylelint-config-wikimedia"
+}
```
stephanebisson, line 3: missing new line

stephanebisson: (Can we have a rule for that?)

hueitan: stylelint already support that, so we are fine with less file in the new line rule.

PR #18, opened 2020-07-01
T250797: Touch interactions and positioning

medied:
Phab ticket: https://phabricator.wikimedia.org/T250797

Required interactions:

- Users can view previews by tapping on links present in an article. 
- After tapping preview is in focus and rest of the screen area has an overlay.
- Users can close the previews either by tapping outside or tap on the close button.

Proposed solution is heavily based on prior ideas (https://github.com/wikimedia/wikipedia-preview/commit/715045cd2f960cfe5621950c4d5d498c33f67e78, https://github.com/wikimedia/wikipedia-preview/pull/12/commits/1c4d089401e1c1433251a0e3076233d306069f46).

Note:

* I was able to test touch interaction on browser (with dev tools mobile simulator) in Firefox only. I tested on Chrome and Safari directly on device, see  https://github.com/wikimedia/wikipedia-preview/pull/18#pullrequestreview-442861832 for more info

---

TODO: 

- [x] Straighten out the `events`
- [x] Close on click outside of popup
- [x] Back screen background
- [x] ~Opportunities for simplification/refactoring~— see https://github.com/wikimedia/wikipedia-preview/pull/18#discussion_r448331162 for more info
- [x] ~Cross browser testing~— see https://github.com/wikimedia/wikipedia-preview/pull/18#issuecomment-655045235


src/utils.js
```diff
@@ -0,0 +1,4 @@
+export const isTouch = () => {
+  // TODO - add IE support
+  return 'ontouchstart' in window
+}
```
hueitan, line 4: This doesn't need to be a function, can be just a constant, as this is not going to change after the page load.

src/touchPopup.js
```diff
@@ -0,0 +1,45 @@
...
+		}
+		if ( events.onHide ) {
+			popupEvents.onHide = events.onHide
+		}
+	}
+
+	return { show, hide, subscribe }
+}
+
+export { createTouchPopup }
```
hueitan, line 45: Can we have a common popup component, then touchable popup and desktop popup can enhance using the common popup component. so most parts of it are the same.

medied: Yea I was thinking along the same lines actually (hence "Opportunities for simplification/refactoring" TODO in PR the description). As a starting point I'm using what Stephane [shared earlier](https://github.com/wikimedia/wikipedia-preview/pull/12#issuecomment-648900604):

>...when we do mobile, we can update "Popup" to handle it or, even better, create another container that acts as a bottom drawer and let the index decide which container to use depending on the platform.

And that line of thinking was already reflected in https://github.com/wikimedia/wikipedia-preview/commit/715045cd2f960cfe5621950c4d5d498c33f67e78 actually. Looking at the codebase now, it seems there's definitely some opportunity for simplification. Already seeing that `show` and `hide` will do different things depending on popup though, so we're probably going to have to make use of `isTouch` in there.

src/utils.js
medied: Good call, will update in upcoming commit along with making sure it works for all browsers. Currently only working in Firefox. Might need to add something like:

```
if (('ontouchstart' in window) ||
     (navigator.maxTouchPoints > 0) ||
     (navigator.msMaxTouchPoints > 0)) {
      /* browser with either Touch Events of Pointer Events
         running on touch-capable device */
}
```
https://hacks.mozilla.org/2013/04/detecting-touch-its-the-why-not-the-how/

style/preview.less
```diff
@@ -143,4 +143,18 @@
...
     }
+
+    &.mobile {
+        width: 100%;
+    }
+}
+
+.wp-dark-screen {
+    background-color: rgba(0,0,0, 0.7);
+    z-index: 100; // TODO make it a variable?
```
medied, line 154: What do you think about adding a LESS variable for the z-index of the popup? That way here we could set the z-index of the screen to `@popup-z-index - 10` or something

src/touchPopup.js
medied: What about something like https://github.com/wikimedia/wikipedia-preview/commit/43dc2c1dceb379181f8b28afcc9b07d12ebf4c7a. Sharing in a exploratory sub-branch just in case before committing to official branch, highlights:

* Abstract the common pieces to popup.js, leave the touch/hover specific code to hoverPopup and touchPopup respectively. 
* Simplify styles by getting rid of touchpopup.less altogether and including variable for `popup` z-index along with all popups styles directly in popup.less 
* Have `isTouch` return a variable instead of function

hueitan: > * Simplify styles by getting rid of touchpopup.less altogether and including variable for `popup` z-index along with all popups styles directly in popup.less
> * Have `isTouch` return a variable instead of function

Let's take these two into this branch.

> * Abstract the common pieces to popup.js, leave the touch/hover specific code to hoverPopup and touchPopup respectively.

Although I said this simplication/refactoring in the beginning, can we wait and see until more components (ex. gallery), so we have the full picture to see how we do the refactoring here.

package.json
```diff
@@ -11,7 +11,7 @@
     "build:dev": "webpack --mode development",
     "build:prod": "webpack --mode production",
     "build": "npm run build:dev && npm run build:prod",
-    "dev": "webpack-dev-server --mode development"
+    "dev": "webpack-dev-server --mode development --host=0.0.0.0"
```
hueitan, line 14: why the host flag?

src/touchPopup.js
medied: Yea that sounds good. It's why I shared in a different branch, the refactoring would be somewhat significant

package.json
medied: It's for mobile local testing:

https://stackoverflow.com/questions/35412137/how-to-get-access-to-webpack-dev-server-from-devices-in-local-network

I was able to test/dev with the dev tools mobile emulator in Firefox without any issues, but Chrome and Safari were a different story. I couldn't get touch events working on those browsers ([one](https://stackoverflow.com/questions/53804541/activate-touch-events-on-chrome-71), [two](https://github.com/fabricjs/fabric.js/issues/5362)). We can however take advantage of this method to test local environment directly on your device: 

* with the `--host` flag there, just `npm run dev` 
*  then make sure both device and laptop are connected to the same wifi
* on device, navigate to the correct IP ([android](https://stackoverflow.com/questions/4779963/how-can-i-access-my-localhost-from-my-android-device), [iphone](https://stackoverflow.com/questions/3132105/how-do-you-access-a-website-running-on-localhost-from-iphone-browser)) and then we can easily test

This is how I was able to test on Chrome and Safari. I'm still taking care of the CBT account, so not sure how it would perform there.

style/preview.less
```diff
@@ -143,4 +143,8 @@
 			}
 		}
 	}
+
+	&.mobile {
+		width: 100%;
+	}
```
hueitan, line 149: In the layout ticket https://phabricator.wikimedia.org/T250796

> At max, **it expands up to 80% of the screen height** and if there is still more content is available to cover then stick preview header(thumbnail, Wikipedia wordmark, close icon) and enable vertical scroll. It ends with a CTA to read the entire article by visiting mobile Wikiped

Currently the height is set by `max-height` with the fixed pixels, I would expect we apply `calc( 80wh )` here to actually get max 80% of the height size when expanded.

src/event.js
```diff
@@ -11,12 +12,24 @@ export const customEvents = popup => {
 		onExpand = () => {
 			popup.element.component.wikipediapreviews.classList.add( 'expanded' )
 		},
+
+		onTouchStart = e => {
+			const targetClassName = e.target.parentNode.className
+			if ( targetClassName.indexOf( 'wikipediapreviews-' ) === -1 ) {
+				popup.hide()
+			}
```
hueitan, line 20: one case found, in popup mode, touch the text in `bold`, for example cat article, touch `cat` `domestic cat` `house cat` `farm cat` `feral cat` will close the popup, but it shouldn't.

medied: Good catch 👀

hueitan: use the similar code in `onMouseLeave` event should work to cover all elements in the popup

```js
!popup.element.contains( toElement )
```

style/preview.less
hueitan: opps, I think this is out of the scope of this ticket. 

This should be part of the Layout task - let's work on another PR for this 

1. 80% screen height
1. visiting mobile Wikipedia; visiting desktop Wikipedia based on `isTouch`
1. vertical scrolling in expanded mode

src/event.js
medied: Done

medied:
Should we add a screenshot of the touch/mobile version of the popup in the README next to the hover one?

Something like:

Mobile | Desktop
--- | ---
 | screenshot

hueitan:
> Should we add a screenshot of the touch/mobile version of the popup in the README next to the hover one?
> 

Yeah I like this table idea

hueitan:
We can ignore e2e cross browser testing test spec right now, and one last time is adding unit test for touchPopup, that can be similar to popup component + background dark screen visible/invisible .

test/touchPopup.test.js
```diff
@@ -0,0 +1,54 @@
...
+
+	it( 'hides the popup and removes background screen when hide event is triggered', () => {
+		popup.hide()
+		assert.equal( popupElement.style.visibility, 'hidden' )
+    assert( onHideCallback.called )
+    
+    const backgroundScreen = dom.window.document.querySelector('.wp-dark-screen')
+    assert.ifError(backgroundScreen)
+  } )
+} )
```
hueitan, line 54: remember to run npm run test:fix-lint

package.json
stephanebisson: I did it by attaching the `touchstart` event to the root back then, not sure why, but it works in Chrome

https://github.com/wikimedia/wikipedia-preview/blob/T240006-mobile/src/index.js#L29

style/preview.less
stephanebisson: The rules above should be part of this task. Can be a separate PR if you wish. In which case you should bring the task back to "dev".

The layout task is about the content of the preview. This task is about where it goes on mobile including sizing.

medied: I was literally about to create a separate PR for it (https://github.com/wikimedia/wikipedia-preview/tree/T250796-layout-touch) and was wondering if it should point to this PR or master...  so this simplifies things! Let me merge with this PR and push changes now then

hueitan: 6519665F-4BAF-4C08-86A8-43032F5BD43A
66B87BBA-81C2-4893-98CB-DB8F85BE4296

src/touchPopup.js
```diff
@@ -6,11 +6,13 @@ const addBackgroundScreen = ( document ) => {
 		const screen = document.createElement( 'div' )
 		screen.classList.add( 'wp-dark-screen' )
 		document.body.appendChild( screen )
+		document.body.style.overflow = 'hidden'
 	},
 
 	removeBackgroundScreen = ( document ) => {
 		const screen = document.getElementsByClassName( 'wp-dark-screen' )
 		document.body.removeChild( screen[ 0 ] )
+		document.body.style.overflow = 'scroll'
```
hueitan, line 15: the client might have their own overflow, we need to rollback to the original overflow if there is given.

PR #20, opened 2020-07-09
T256213: Add a small delay to onMouseLeave

medied:
Phab ticket: https://phabricator.wikimedia.org/T256213

What's needed:

> add a small delay to the closing of the popup when the mouse leaves the word

We tested a few delays with [this demo branch](https://github.com/wikimedia/wikipedia-preview/commit/0b6eaf28d8aeeb6a7b15c0630e7473597d71f8a6) and we're setting delay to 300ms based on that.

### DIFF
--- a/src/event.js
+++ b/src/event.js
@@ -5,7 +5,7 @@ export const customEvents = popup => {
 			const toElement = e.toElement || e.relatedTarget || e.target
 			if ( toElement !== popup.element.currentTargetElement &&
 			!popup.element.contains( toElement ) ) {
-				popup.hide()
+				setTimeout( popup.hide, 300 )
 			}
 		},
 
--- a/src/event.js
+++ b/src/event.js
@@ -8,18 +8,20 @@ export const customEvents = popup => {
 			if ( toElement !== previewElement && !popup.element.contains( toElement ) ) {
 				let timeoutId
 				const persistPopup = () => {
+						popup.element.removeEventListener( 'mouseenter', persistPopup )
+						popup.timeoutId = null
 						clearTimeout( timeoutId )
 					},
 
 					hidePopup = () => {
 						popup.element.removeEventListener( 'mouseenter', persistPopup )
-						previewElement.removeEventListener( 'mouseenter', persistPopup )
+						popup.timeoutId = null
 						popup.hide()
 					}
 
 				timeoutId = setTimeout( hidePopup, 300 )
 				popup.element.addEventListener( 'mouseenter', persistPopup )
-				previewElement.addEventListener( 'mouseenter', persistPopup )
+				popup.timeoutId = timeoutId
 			}
 		},
 
--- a/src/event.js
+++ b/src/event.js
@@ -2,13 +2,12 @@ import { isTouch } from './utils'
 
 export const customEvents = popup => {
 	const onMouseLeave = e => {
-			const toElement = e.toElement || e.relatedTarget || e.target
-			if ( toElement !== popup.element.currentTargetElement &&
-			!popup.element.contains( toElement ) ) {
-				let timeoutId
-				const previewElement = popup.element.currentTargetElement,
+			const toElement = e.toElement || e.relatedTarget || e.target,
+				previewElement = popup.element.currentTargetElement
 
-					persistPopup = () => {
+			if ( toElement !== previewElement && !popup.element.contains( toElement ) ) {
+				let timeoutId
+				const persistPopup = () => {
 						clearTimeout( timeoutId )
 					},
 
--- a/src/event.js
+++ b/src/event.js
@@ -5,7 +5,22 @@ export const customEvents = popup => {
 			const toElement = e.toElement || e.relatedTarget || e.target
 			if ( toElement !== popup.element.currentTargetElement &&
 			!popup.element.contains( toElement ) ) {
-				setTimeout( popup.hide, 300 )
+				let timeoutId
+				const previewElement = popup.element.currentTargetElement,
+
+					persistPopup = () => {
+						clearTimeout( timeoutId )
+					},
+
+					hidePopup = () => {
+						popup.element.removeEventListener( 'mouseenter', persistPopup )
+						previewElement.removeEventListener( 'mouseenter', persistPopup )
+						popup.hide()
+					}
+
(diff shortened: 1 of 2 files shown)


medied:
I had slightly misunderstood the ticket, you're right. How about https://github.com/wikimedia/wikipedia-preview/pull/20/commits/9c2f0633ac609fbafb24e1eb22affd041f565654? The more we test it, the better

src/event.js
```diff
@@ -5,7 +5,22 @@ export const customEvents = popup => {
 			const toElement = e.toElement || e.relatedTarget || e.target
 			if ( toElement !== popup.element.currentTargetElement &&
 			!popup.element.contains( toElement ) ) {
-				popup.hide()
+				let timeoutId
+				const previewElement = popup.element.currentTargetElement,
```
hueitan, line 9: `popup.element.currentTargetElement` used twice within the same scope, let's define `previewElement` in the top of the scope.

jpita:
Tested on android, windows (IE, Chrome and FF) and mac.
Found this small issue when 2 words are together.
When I move from one word immediately to the other, the popup on the 2nd word closes even though the mouse is over the 2nd word.
It seems we are not detecting the 2nd mouse-enter and stopping the timer.
Jul-14-2020 15-41-43

hueitan:
> Tested on android, windows (IE, Chrome and FF) and mac.
> Found this small issue when 2 words are together.
> When I move from one word immediately to the other, the popup on the 2nd word closes even though the mouse is over the 2nd word.
> It seems we are not detecting the 2nd mouse-enter and stopping the timer.

🥇 for u @jpita!!!! I can confirm this behaviour and we need to fix it.

hueitan:
Following what Pita found, what if when user mouse leaves the popup/`previewElement` to another wikipedia hyperlinks, then mouse move back to the popup within 300ms, do you show the new popup or keep the current popup? I think the latter. If so our code need to do extra guard here.

src/index.js
```diff
@@ -16,8 +16,14 @@ function init( {
 			createPopup( popupContainer ),
 		events = customEvents( popup ),
 		showPopup = ( { target } ) => {
+			if ( popup.timeoutId ) {
+				clearTimeout( popup.timeoutId )
+				popup.timeoutId = null
+			}
```
hueitan, line 22: it works basically but it doesn't clear the event listener through `hidePopup`. 

We can check whether the popup exists, then call the event `hidePopup` before showing the new popup, in this case we don't need to expose `timeoutId` as the attribute of popup.

jpita: still not ready to test, right?
when it is please move it to QA and lmk?

hueitan:
note : to review this https://github.com/wikimedia/wikipedia-preview/commit/beb4a3a7ca740947214c03e7ee11982697c775a1

jpita:
this still has pending changes, right?

stephanebisson:
Works well as far as I can tell but we need unit tests to verify the different scenarios: exit, exit and re-enter the same link or popup, exit and re-enter a different link.

medied:
Yes as of https://github.com/wikimedia/wikipedia-preview/pull/20/commits/e1e18a4990293db3c19b3387dc342275247e6cdb the expected behavior is overall working well but under the hood there are some event listeners we are still failing to remove. To address this, we are in the process of merging #23 into this PR. 

Also, as noted in #23, when you are QAing this PR @jpita keep in mind that transitioning from a preview in one language to another one in a different language (all within 300ms) you will see a buggy behavior. We plan to address this in a new ticket

src/index.js
```diff
@@ -16,7 +16,9 @@ function init( {
 			createPopup( popupContainer ),
 		events = customEvents( popup ),
 		showPopup = ( { target } ) => {
-			popup.hide()
+			if ( popup.element.style.visibility === 'visible' ) {
```
medied, line 19: @hueitan just FYI: I realized tonight that always executing `popup.hide()` without a check was breaking the previews in mobile with the error below (it tries to remove a background screen that's not there), so I'm including this conditional

```
Uncaught TypeError: Node.removeChild: Argument 1 is not an object.
    removeBackgroundScreen webpack://wikipediaPreview/./src/touchPopup.js?:17
    hide webpack://wikipediaPreview/./src/touchPopup.js?:49
    showPopup webpack://wikipediaPreview/./src/index.js?:31
```

medied:
> ...but we need unit tests to verify the different scenarios

Huei and I suspected these tests would be somewhat tricky/particular, I gave it a first try today and indeed ran into issues on how to approach them and think about measuring timing. I'm looking to pick someone's brain to understand what makes most sense. For what it's worth, this sub-branch shows what I tried: [`T256213-hover-closing-delay-tests-exploration`](https://github.com/wikimedia/wikipedia-preview/commit/683e7caf043722847009bdb1b98447da8b4f23e5).

@stephanebisson would you say the tests here are a merge blocker?

Either way, @jpita I'd say you can now start QAing this ticket. Just don't merge to master until we settle on tests

jpita:
> Either way, @jpita I'd say you can now start QAing this ticket. Just don't merge to master until we settle on tests

will move the task to code review until the code is finished so it's not stuck in QA

hueitan:
> Huei and I suspected these tests would be somewhat tricky/particular, I gave it a first try today and indeed ran into issues on how to approach them and think about measuring timing. I'm looking to pick someone's brain to understand what makes most sense. For what it's worth, this sub-branch shows what I tried: [`T256213-hover-closing-delay-tests-exploration`](https://github.com/wikimedia/wikipedia-preview/commit/683e7caf043722847009bdb1b98447da8b4f23e5).
> 
> @stephanebisson would you say the tests here are a merge blocker?

This is really tricky, this approach is going into an integration test which is not a unit test anymore.

I would suggest don't let this be the merge blocker, the unit test (only event.js) or e2e test (integration) should cover this scenario + this as well https://github.com/wikimedia/wikipedia-preview/pull/23#issuecomment-661379936,

jpita:
I can see what I can do with cypress

hueitan:
We should move on, @jpita, currently there is 0% in the integration test, let's qa this then feel free to add them later.

jpita:
ok, this is tested, feel free to merge when you think is ready

PR #21, opened 2020-07-10
T250799: Image gallery within Wikipedia Preview

medied:
Phab ticket: https://phabricator.wikimedia.org/T250799

Add gallery with required interactions:

* When a reader swipes up, they can see multiple images from an article in a gallery view
* They can view each image independently
* Images are a gallery of pictures from the article which is being previewed
* Gallery does not include video and audio content
* If an article does not have images, no image gallery options will show

**PR highlights**:
* Methods for mini gallery row and full screen gallery in `gallery.js`
* `requestPageMedia` and `renderPreviewMedia` to handle gallery-related logic separately
  * This lets us only load images as necessary for faster popup rendering 
---

TODOs

- [x] add full-screen gallery mode
- [x] refine styling 
- [x] review data load order
- [x] add unit tests
- [x] ~add licensing info~ will be addressed in separate ticket


.eslintrc.js
```diff
@@ -16,6 +16,7 @@ module.exports = {
     sourceType: 'module'
   },
   rules: {
-    semi: [2, "never"]
+    semi: [2, "never"],
+    'max-statements-per-line': 'off'
```
medied, line 20: Without updating this rule we get this linter error in `getGalleryRow`:

```
This line has 2 statements. Maximum allowed is 1  max-statements-per-line
```
I believe it's also an option to set it to 1 for a warning instead of an error, if we don't want to turn it off completely

src/gallery.js
```diff
@@ -0,0 +1,103 @@
...
+	getGalleryRow = ( mediaItems ) => {
+		const galleryRow = document.createElement( 'div' )
+		galleryRow.classList.add( 'wikipediapreview-gallery-row' )
+
+		mediaItems.forEach( item => {
+			const image = document.createElement( 'img' )
+
+			image.classList.add( 'wikipediapreview-gallery-image' )
+			image.src = item.src
+			image.addEventListener( 'click', ( e ) => { showFullscreenGallery( e, mediaItems ) } )
```
hueitan, line 96:
```suggestion
			image.addEventListener( 'click', ( e ) => { 
			    showFullscreenGallery( e, mediaItems ) 
			} )
```

This will fix the eslint error

src/index.js
```diff
@@ -19,10 +19,20 @@ function init( {
...
 					popup.show( renderPreview( lang, data, isTouch ), target )
 				}
+
+				popup.lang = lang
+				popup.title = title
+				const expanded = document.querySelector( '.wikipediapreview.expanded.mobile' )
+				if ( expanded ) {
+					requestPageMedia( lang, title, mediaData => {
+						renderPreviewMedia( document, mediaData )
+					} )
```
hueitan, line 34: moving this into the event or util, so we can reuse in both index.js and event.js

src/gallery.js
```diff
@@ -0,0 +1,103 @@
...
+
+	getGalleryRow = ( mediaItems ) => {
+		const galleryRow = document.createElement( 'div' )
+		galleryRow.classList.add( 'wikipediapreview-gallery-row' )
+
+		mediaItems.forEach( item => {
+			const image = document.createElement( 'img' )
+
+			image.classList.add( 'wikipediapreview-gallery-image' )
+			image.src = item.src
```
hueitan, line 95: i would suggest adding the image as the background image and update the background position / cover, otherwise there is stretch on the image

image

src/gallery.js
```diff
@@ -0,0 +1,103 @@
...
+			image.addEventListener( 'click', ( e ) => { showFullscreenGallery( e, mediaItems ) } )
+			galleryRow.appendChild( image )
+		} )
+
+		return galleryRow
+	}
+
+export { getGalleryRow, showFullscreenGallery }
```
hueitan, line 103: it is interesting to see we have a different approach here to build the view.

**In preview and popup**, we use the way of the template (similar to Backbone .template file), therefore preview.js keeps only the html and the event being stored in the event.js

**In gallery**, we use the fundamental way of building the view + event, therefore the gallery is fully controlled in this way, which also brings some benefit but maintenance cost.

---

These two methods can become harder for us to maintain in the future, and also we don't use any framework. Anyway, this doesn't harm as we are in the prototype stage, do keep this in our mind for future reference.

jpita:
@medied 
* when I reach the first/last image on the gallery, the arrow does not disappear (it should since the tap does nothing).
* we should have an area to tap on instead of just the arrow to make it easy for the user to flip images
image
* landscape mode is not working very well, the image is cropped and the arrow buttons are almost hidden
image

src/gallery.js
medied: Good call, this way it matches better with thumbnail in header as well

medied:
> * when I reach the first/last image on the gallery, the arrow does not disappear (it should since the tap does nothing).
> * we should have an area to tap on instead of just the arrow to make it easy for the user to flip images
> * landscape mode is not working very well, the image is cropped and the arrow buttons are almost hidden

Thanks Pita! Am discussing these with Sudhanshu now to likely address in separate tickets

medied:
> Unit test coverage is significantly decreased in this PR. Please have a look.

Okay so now we have tests for the main new methods introduced in this PR:

* `requestPageMedia` in api.test.js
* `showFullscreenGallery` and `getGalleryRow` in brand new gallery.test.js
* `renderPreviewMedia` in preview.test.js

Anything else we're missing? 🤔

src/utils.js
```diff
@@ -16,5 +18,12 @@ export const getWikipediaAttrFromUrl = url => {
...
 		return null
 	},
+
 	isTouch = 'ontouchstart' in window || ( navigator.maxTouchPoints > 0 ) ||
-		( navigator.msMaxTouchPoints > 0 )
+		( navigator.msMaxTouchPoints > 0 ),
+
+	addMiniGalleryRow = ( lang, title ) => {
+		requestPageMedia( lang, title, mediaData => {
+			renderPreviewMedia( document, mediaData )
```
hueitan, line 27: missing case, when there is no mediaData (empty array), we don't need to render the gallery.

jpita:
can I merge this or there is still pending changes?

medied:
> can I merge this or there is still pending changes?

Give me one moment

src/utils.js
```diff
@@ -24,6 +24,8 @@ export const getWikipediaAttrFromUrl = url => {
 
 	addMiniGalleryRow = ( lang, title ) => {
 		requestPageMedia( lang, title, mediaData => {
-			renderPreviewMedia( document, mediaData )
+			if ( mediaData ) {
```
hueitan, line 27: the response mediaData can be `[]` empty array
```suggestion
			if ( mediaData && mediaData.length > 1 ) {
```

medied: You are correct sir, just tested with an article with no images

jpita:
this is tested.
feel free to merge when u think is ready.
if there are more commits I'll need to test it again

stephanebisson:
@medied I'll do final review and testing after the merge conflict is resolved.

medied:
Okay how about https://github.com/wikimedia/wikipedia-preview/pull/21/commits/b5850e810916bc892cb051d91606b4088ffd5577:

* Appends the fullscreen gallery to `.wikipediapreview-gallery` container, rather than directly appending to `body`
* Makes all gallery-related functions operate on a popup instance rather than `document`
* Simplifies code a bit by actually getting rid of `renderMediaPreview` and `addMiniGallery` altogether, effectively leaving `getGalleryRow` as the pure function that receives the media data (and `popup`) and returns the HTML representation

src/index.js
```diff
@@ -28,6 +29,19 @@ function init( {
 				if ( data ) {
 					popup.show( renderPreview( lang, data, isTouch ), target )
 				}
+
+				popup.lang = lang
+				popup.title = title
+				const expanded = root.querySelector( '.wikipediapreview.expanded.mobile' )
+
+				if ( expanded ) {
+					requestPageMedia( lang, title, mediaData => {
```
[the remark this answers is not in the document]
medied: Two cases depending on whether the preview text is long enough. If it's short enough and the gallery row fits then we `requestPageMedia` in index.js. Otherwise when user clicks continue reading CTA and expands the preview, we `requestPageMedia` in event.js.

Huei first noticed it [here](url) and that's how we had [ended up](https://github.com/wikimedia/wikipedia-preview/pull/21/commits/e05409353497ae51533f28562f38493256a61bf6) with `addMiniGalleryRow` in utils

Re-reading his comment again now I suppose we could make those few lines available through `events` when `popup` does `subscribe` if we want to make it more DRY

src/index.js
```diff
@@ -28,6 +29,19 @@ function init( {
...
+						if ( mediaData && mediaData.length > 1 ) {
```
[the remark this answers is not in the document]
medied: Good catch! I also removed the previous/next buttons in this case

src/index.js
```diff
@@ -28,6 +29,19 @@ function init( {
...
 					popup.show( renderPreview( lang, data, isTouch ), target )
 				}
+
+				popup.lang = lang
+				popup.title = title
+				const expanded = root.querySelector( '.wikipediapreview.expanded.mobile' )
+
+				if ( expanded ) {
+					requestPageMedia( lang, title, mediaData => {
+						if ( mediaData && mediaData.length > 0 ) {
```
stephanebisson, line 39: You have `length > 0` here and `length > 1` in `event.js`. That's a good enough reason to make it more DRY.

src/gallery.js
```diff
@@ -0,0 +1,122 @@
...
+			renderNext( galleryContainer )
+		} )
+		previousButton.classList.add( 'wp-gallery-popup-button', 'previous' )
+		previousButton.style.opacity = current === 0 ? '0.5' : '1'
+		previousButton.addEventListener( 'click', () => {
+			renderPrevious( galleryContainer )
+		} )
+
+		if ( gallery.length > 1 ) {
+			Array.prototype.forEach.call(
```
stephanebisson, line 84: Why do you use this syntax instead of `[...].forEach(...)`?

src/index.js
medied: Ha well there you go... do you have a preference for making it available through `utils.js` or `events.js`?

src/gallery.js
medied: Hmm I thought there was a browser compatibility issue but apparently not, `[...].forEach(...)` seems to be working fine and it's more readable. Updating...

src/index.js
medied: This seems to be more in line with the code: https://github.com/wikimedia/wikipedia-preview/pull/21/commits/04ae5e0b6bc26d88b94421bac90b1f95696506ad

src/index.js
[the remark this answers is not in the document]
hueitan: what a mistake from me 😝 thanks for checking this detail @stephanebisson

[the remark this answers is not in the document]
hueitan: even better this can short to

```diff
- if ( mediaData && mediaData.length > 0 ) {
+ if ( mediaData && mediaData.length ) {
```

PR #23, opened 2020-07-17
stack timeout listener

hueitan:
PR related to https://github.com/wikimedia/wikipedia-preview/pull/20 
 Comment related to https://github.com/wikimedia/wikipedia-preview/commit/beb4a3a7ca740947214c03e7ee11982697c775a1#commitcomment-40695177

Before we introducing the **delay feature for desktop**, we deal with all the events synchronously, **delay feature** brings the race condition and asynchronous event execution. Also note that there is only one instance of popup, unless we declare popup instance for each new popup event.

Therefore this let us think more about **the order of the event called**, and all of them should be **cleared empty once hide event is called**, this POC shows the following

1. A central place for all the event listener - No more manually remove each event listener, instead, remove ALL
1. A central place for all the timeout - takes care of the timeout event whether it should execute or clear.

Let me know what do you see or foresee any problem.

### DIFF
--- a/src/event.js
+++ b/src/event.js
@@ -2,7 +2,7 @@ import { isTouch } from './utils'
 
 export const customEvents = popup => {
 
-	var eventListenerStack = [],
+	let eventListenerStack = [],
 		timeoutStack = []
 	const addTimeout = ( func, timeout ) => {
 			const id = setTimeout( func, timeout )
--- a/src/event.js
+++ b/src/event.js
@@ -19,7 +19,7 @@ export const customEvents = popup => {
 			target.addEventListener( type, listener, options )
 			eventListenerStack.push( [ target, type, listener, options ] )
 		},
-		clearEventListener = () => {
+		clearAllEventListener = () => {
 			eventListenerStack.forEach( eventListener => {
 				const [ target, type, listener, options ] = eventListener
 				target.removeEventListener( type, listener, options )
@@ -53,7 +53,7 @@ export const customEvents = popup => {
 		},
 
 		onHide = () => {
-			clearEventListener()
+			clearAllEventListener()
 			clearAllTimeout()
 		},
 
--- a/src/event.js
+++ b/src/event.js
@@ -1,27 +1,43 @@
 import { isTouch } from './utils'
 
 export const customEvents = popup => {
-	const onMouseLeave = e => {
+
+	var eventListenerStack = [],
+		timeoutStack = []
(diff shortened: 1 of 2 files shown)


medied:
I really like this approach Huei, it's a neat way to keep track of all events and therefore easily remove them. 

I think we should move forward in this direction. However one issue I'm just realizing now is that a new popup will still accidentally hide if you go from one preview from one init instance/section/language to a different one:

```
<script type="text/javascript">
      wikipediaPreview.init({
        lang: 'en',
        root: document.querySelector('.content-en')
      });
      wikipediaPreview.init({
        lang: 'fr',
        root: document.querySelector('.content-fr')
      });
      wikipediaPreview.init({
        lang: 'es',
        root: document.querySelector('.content-es')
      });
      wikipediaPreview.init({
        lang: 'zh',
        root: document.querySelector('.content-zh')
      });
</script>
```

It seems a new popup (from a different `init` instance) activated within the 300ms does not have the updated stacks from the previous popup (of different `init` instance). I tried making both stacks somehow more globally accessible (using `localStorage`, declaring stacks in index.js and importing to event.js, attaching to popup, etc), but can't quite get it to work yet. Will continue trying.

hueitan:
> It seems a new popup (from a different `init` instance) activated within the 300ms does not have the updated stacks from the previous popup (of different `init` instance). I tried making both stacks somehow more globally accessible (using `localStorage`, declaring stacks in index.js and importing to event.js, attaching to popup, etc), but can't quite get it to work yet. Will continue trying.

yup you are right, every single call of `init` has the popup within the scope, therefore my statement before wasn't right at all

> there is only one instance of popup, unless we declare popup instance for each new popup event.

Will look into that! Thanks.

hueitan:
> I think we should move forward in this direction. However one issue I'm just realizing now is that a new popup will still accidentally hide if you go from one preview from one init instance/section/language to a different one:

As we discuss, we will move this into another ticket, as this requires more changes to the overall codebase.

src/event.js
```diff
@@ -1,27 +1,43 @@
 import { isTouch } from './utils'
 
 export const customEvents = popup => {
-	const onMouseLeave = e => {
+
+	var eventListenerStack = [],
```
medied, line 5: `let` instead of `var`?

hueitan: errr, right i don't know why it happens to `var` here, on it

PR #27, opened 2020-07-27
IE: use bounding rectangle left/top instead of x/y

stephanebisson:
https://phabricator.wikimedia.org/T256724

In IE 11, the return value of `element.getBoundingClientRect()` doesn't contain `x` or `y` so the popup positioning was wrong (always below and to the right of the target element). This can cause the document width to change, a scrollbar to appear, and the text to reflow.

I think this was causing the issue described in the task but I can't be sure since I cannot reproduce it anymore.

Note: the change to the `dev` command in `package.json` is needed to allow a local connection to be established with crossbrowsertesting.com

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "build:dev": "webpack --mode development",
     "build:prod": "webpack --mode production",
     "build": "npm run build:dev && npm run build:prod",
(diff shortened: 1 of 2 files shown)


stephanebisson:
This change is needed for compatibility whether or not it fixes the specific bug reported in the task.

I'm merging now but it still needs to go through QA.

PR #28, opened 2020-07-28
T250800 Setup Translatewiki

hueitan:
Phabricator Ticket : https://phabricator.wikimedia.org/T250800

 - [x] Adding Translatewiki needed folder and files
 - [x] Scripting our own i18n utils, and make it flexible enough to migrate banana when needed
    - [x] Feature needed : Proper language fallback
    - [x] Feature needed : Return the key when no found
 - [x] <s>Setup webpack to include json files as an object</s> webpack includes `json-loader` by default

### DIFF
--- a/src/i18n.js
+++ b/src/i18n.js
@@ -1,14 +1,20 @@
 import fallbackMessage from '../i18n/en.json'
-const msg = ( lang, key ) => {
-	let messages
 
-	try {
-		messages = require( `../i18n/${lang}.json` )
-	} catch ( error ) {
-		// Translation not available, discard
+const messages = {
+		en: fallbackMessage
+	},
+
+	msg = ( lang, key ) => {
+		try {
+			if ( !messages[ lang ] ) {
+				messages[ lang ] = require( `../i18n/${lang}.json` )
+			}
+		} catch ( error ) {
+			// Translation not available
+		}
+
+		let message = messages[ lang ]
+		return ( message && message[ key ] ) || fallbackMessage[ key ] || key
 	}
-
-	return ( messages && messages[ key ] ) || fallbackMessage[ key ] || key
-}
 
 export { msg }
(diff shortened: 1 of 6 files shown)


src/i18n.js
```diff
@@ -1,17 +1,14 @@
...
-		'wikipedia': 'Wikipédia'
-	},
-	es: {
-		'wikipedia': 'Wikipedia'
+import fallbackMessage from '../i18n/en.json'
+const msg = ( lang, key ) => {
+	let messages
+
+	try {
+		messages = require( `../i18n/${lang}.json` )
```
stephanebisson, line 6: Is this going to load the file every time the `msg` function is called?

hueitan: nope, the file (object) is being stored within the internal code.

hueitan: sorry, got the wrong understanding here, true, we should make it available once the require is called for the next call used.

hueitan: done, and please check it again.

src/i18n.js
```diff
@@ -1,17 +1,20 @@
...
+	msg = ( lang, key ) => {
+		try {
+			if ( !messages[ lang ] ) {
+				messages[ lang ] = require( `../i18n/${lang}.json` )
+			}
+		} catch ( error ) {
+			// Translation not available
+		}
+
+		let message = messages[ lang ]
```
stephanebisson, line 16: can be `const`

src/i18n.js
```diff
@@ -1,17 +1,20 @@
...
-		'wikipedia': 'Wikipedia'
-	}
-}
 
-const msg = (lang, key) => {
-	return (messages[lang] || messages.en)[key] || key
-}
+	msg = ( lang, key ) => {
+		try {
+			if ( !messages[ lang ] ) {
```
stephanebisson, line 9: I would suggest reworking this logic to not retry to load a missing language all the time. We could put the catch around the `require` and maybe initialize the messages with an empty object when it fails.

hueitan: update, thanks it looks better now.

PR #30, opened 2020-07-29
T250799 gallery followup

medied:
https://phabricator.wikimedia.org/T250799#6339885

Follow up to address Sudhanshu's feedback specifically:

>01. A blue rectangle appears every time while tapping on a forward/backward arrow inside the media viewer. This could be a device-specific thing as I did see this behavior on Samsung A50 but not on my other android device.
>02. Portrait images are appearing blur, it does look good on the Wikipedia mobile web and android app on the same Wifi network.
>03. As discussed in one of our catchups, I think we need to decide the maximum area of display for e.g. check above the mobile web image and how it has left space to accommodate licensing information. As we do this, I would also recommend moving the close icon on the right top side, as of now it's position is kind of looking off.
>04. Sizes of gallery items are smaller(65x62) than what we mentioned in design specs(100,80), is there any challenge to use them as per the suggested size?

Highlevel remarks:

* Includes a high-resolution url for every image 
* Updates/restructures fullscreen gallery HTML
* Updates respective tests

Note: I don't know why this follow-up gallery PR is showing the commits from the original gallery PR under 'Commits' tab, but 'Files changed' tab correctly shows the diff with master


src/gallery.js
```diff
@@ -40,35 +40,44 @@ const hideFullscreenGallery = ( galleryContainer ) => {
 
 	showFullscreenGallery = ( event, mediaItems, popup, doc = document ) => {
 		const fullscreenGallery = doc.createElement( 'div' ),
+			top = doc.createElement( 'div' ),
```
hueitan, line 43: do we need the top container that takes unnecessary 10vh space? 
The closes button element can stay alone without the container, because it has absolute position.

style/popup.less
```diff
@@ -40,11 +37,25 @@
...
+	}
+
+	&-main {
+		display: flex;
+		max-width: 100%;
+		height: 70vh;
+	}
+
+	&-bottom {
+		height: 20vh;
```
hueitan, line 51: I don't find the discussion anywhere, 

1. should the bottom container stay on top of the main? that means main is 100vh.
1. or stay below the main, which the main height should be dynamic based on the size of the caption.

style/popup.less
```diff
@@ -53,36 +64,42 @@
 			background-image: inline-svg( '../images/close-white.svg' );
 			top: 40px;
```
hueitan, line 65: I still think the close button top: 40px is a bit off, the svg is square size,so  top and right position should be the same to looks more correct in general.

src/gallery.js
hueitan: discussion : Let's keep it and let design check in here.

style/popup.less
hueitan: discussion : Let's keep it and let design check in here. or revisit in the image attribution ticket.

style/popup.less
hueitan: discussion : Let's keep it and let design check in here.

PR #31, opened 2020-07-29
T256309: loading state

medied:
https://phabricator.wikimedia.org/T256309

Adds loading state to preview and gallery

---

TODOs

- [x] ~replace preview body .png with the .gif~ animation added instead
- [x] mini gallery row loading
- [x] onError message
- [x] gallery loading


images/loading-thumb.svg
```diff
@@ -0,0 +1,14 @@
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="20" height="20" viewBox="0 0 20 20">
```
stephanebisson, line 1: From now on, please optimize the SVGs you add using [svgo](https://github.com/svg/svgo). It's really easy to do.
```
# install globally so you have it available
npm install -g svgo
# run on the whole image dir or the file you added specifically
svgo images/
svgo images/loading-thumb.svg
```

medied:
A couple of smaller TODOs left, but already open for early feedback

src/loading.js
```diff
@@ -0,0 +1,18 @@
+import '../style/preview.less'
+
+const renderLoading = ( isTouch ) => {
```
medied, line 3: Actually thinking thing of moving `renderLoading` to preview.js, anyone agree that's better?

src/gallery.js
```diff
@@ -1,7 +1,64 @@
...
+		component.classList.add( 'wp-gallery-popup-loading' )
+		spinnerContainer.classList.add( 'wp-gallery-popup-loading-spinner-container' )
+		spinner.classList.add( 'wp-gallery-popup-loading-spinner' )
+		bounceBegin.classList.add( 'wp-gallery-popup-loading-bounce' )
+		bounceEnd.classList.add( 'wp-gallery-popup-loading-bounce' )
+		stillLoading.classList.add( 'wp-gallery-popup-loading-still' )
+		stillLoading.innerHTML = 'Still loading'
+
+		spinner.appendChild( bounceBegin )
+		spinner.appendChild( bounceEnd )
```
medied, line 21: Does anyone else also get a linter error with this?

```
[ bounceBegin, bounceEnd ].forEach( bounce => {
   spinner.appendChild( bounce )
} )
```

```
ERROR in ./src/gallery.js
Module Error (from ./node_modules/eslint-loader/dist/cjs.js):


/Users/emedina/internet/wikimedia/wikipedia-previews/src/gallery.js
  17:4   error  Unexpected newline between object and [ of property access  no-unexpected-multiline
  17:17  error  Unexpected use of comma operator
```

https://eslint.org/docs/2.0.0/rules/no-unexpected-multiline

hueitan:
```diff
- [ bounceBegin, bounceEnd ].forEach( bounce => {
+ const bounceArray = [ bounceBegin, boundEnd ]
+ bounceArray.forEach( bounce => {
   spinner.appendChild( bounce )
} )
```

can fix that

src/loading.js
hueitan: we don't have one general loading behaviour, instead we have the different loading in preview / gallery.

therefore I think we can have the `previewLoading` component, then `preview` component use it.
same to gallery popup loading component.

package.json
```diff
@@ -40,6 +40,7 @@
     "style-loader": "^1.1.3",
     "stylelint-config-wikimedia": "^0.10.1",
     "stylelint-webpack-plugin": "^2.1.0",
+    "url-loader": "^4.1.0",
```
stephanebisson, line 43: Could you point out where this is being used?

src/gallery.js
```diff
@@ -1,7 +1,74 @@
...
+				text.style.visibility = 'hidden'
+				image.style.visibility = 'visible'
+				image.removeEventListener( 'load', onLoad )
+			},
+
+			onError = () => {
+				clearTimeout( timeoutId )
+				error.style.visibility = 'visible'
+				spinner.style.visibility = 'hidden'
+				text.innerHTML = 'There was an error loading this image'
```
stephanebisson, line 56: This string should be localized.

medied:
> Works well but gets stuck in loading state when the response is not supported (type !== 'standard' or dir !== 'ltr'). I guess it could go to the error state in those cases.

Yea it should go to error state in those cases, we have [T257811](https://phabricator.wikimedia.org/T257811) for that. Looking at that ticket now, I'm realizing this PR already covers the gallery error half. Shall we leave the other half (preview error) for T257811?

> Also, I'm becoming uneasy with the proliferation of imperative DOM building (document.createElement, etc). I think it's fine for a simple element like a container or an overlay that is basically just a div with a class but for anything that has a structure with a corresponding css structure, I would rather see it as a string literal.

How about https://github.com/wikimedia/wikipedia-preview/pull/31/commits/f33de5dbed99f76694b5abc2a4122ff8bd55676b? I agree it was starting to get a bit convoluted in gallery.js. I'm already liking this string literal approach more: seems more readable, reduces the number of code lines in gallery.js, decreases the use of `document`

package.json
medied: It's for the body of the preview loading state, which is currently an image (.png) but it's meant be animated (.gif)

https://github.com/wikimedia/wikipedia-preview/blob/T256309-loading-state/style/preview.less#L98

However, I synced briefly with Sudhanshu today and we think it might be better to achieve the animation with CSS rather than a gif, kind of like we are doing with [the spinner](https://github.com/wikimedia/wikipedia-preview/blob/T256309-loading-state/style/popup.less#L180). In that case we wouldn't need `url-loader`. I'm gonna look in the CSS animations now, will keep you updated

medied: Added animation, removed `url-loader`

src/loading.js
medied: Yea looking at this again, `renderLoading` seems sort of lonely in loading.js by itself. I think it makes sense to have it make some company to `renderPreview` in preview.js, they are are both aligned shared some logic

medied:
Does it make sense to add tests for `renderLoading` and `getFullScreenGallery`?

Also, realizing something we can do now to [improve `requestPageMedia`](https://github.com/wikimedia/wikipedia-preview/commit/b24bd8280a8c02ac4f798e7831e960b0631a8470#diff-e1d62d8c4965edf0e532dfa19da0891fR26): when constructing `source` we could do a small check to see if the link correctly ends with valid format (png, jpg, etc... gif?) and if not set `media.src: thumbnail` as a fallback. Does that make sense? The first preview link of the Afrika article has an example of a `source` set incorrectly (the last gallery item, earth gif). It's actually helpful right now to test the gallery loading error state though

src/gallery.js
```diff
@@ -1,19 +1,101 @@
...
+				toggleLoading( loading, image )
+				image.src = gallery[ current ].src
+				refresh.removeEventListener( 'click', onRefresh )
+			},
+
+			onError = () => {
+				clearTimeout( timeoutId )
+				error.style.visibility = 'visible'
+				spinner.style.visibility = 'hidden'
+				text.innerHTML = msg( lang, 'gallery-loading-error' )
```
stephanebisson, line 65: Please use `innerText` as much as possible so it is not vulnerable to HTML injection.

src/gallery.js
```diff
@@ -1,19 +1,101 @@
...
+				text.style[ 'font-size' ] = '16px'
+				text.style.visibility = 'visible'
+				refresh.innerHTML = msg( lang, 'gallery-loading-error-refresh' )
```
stephanebisson, line 68: innerText

src/gallery.js
```diff
@@ -59,50 +135,39 @@ const hideFullscreenGallery = ( galleryContainer ) => {
...
-		fullscreenGallery.classList.add( 'wp-gallery-popup' )
-		top.classList.add( 'wp-gallery-popup-top' )
-		main.classList.add( 'wp-gallery-popup-main' )
-		bottom.classList.add( 'wp-gallery-popup-bottom' )
+		galleryContainer.insertAdjacentHTML( 'beforeend', fullscreenGallery )
+
+		let image = galleryContainer.querySelector( 'img' ),
+			caption = galleryContainer.querySelector( '.wp-gallery-popup-caption' ),
+			loading = galleryContainer.querySelector( '.wp-gallery-popup-loading' ),
+			nextButton = galleryContainer.querySelectorAll( '.wp-gallery-popup-button.next' )[ 0 ],
```
stephanebisson, line 143: Is there a difference between `querySelectorAll(...)[0]` and `querySelector(...)` ?

medied: Hmm not if we query only `next`/`previous`, that's more readable 👍

PR #33, opened 2020-07-29
Localize wordmark with strings

stephanebisson:
https://phabricator.wikimedia.org/T258974

Use a string for the Wikipedia wordmark and let it be localized.

I tried using the 22 known SVGs but it brought the production bundle from 38k to 240k.


i18n/en.json
```diff
@@ -1,4 +1,5 @@
 {
     "continue-reading": "Continue Reading",
-    "read-more": "Read more on Wikipedia"
+    "read-more": "Read more on Wikipedia",
+    "wikipedia-wordmark": "WikipediA"
```
hueitan, line 4: is this a common in the translatewiki community to do the wordmark translation?

stephanebisson: I don't know if it's common but I also don't think it is special. It should only be the localized name of the site the preview is coming from.

stephanebisson:
Handled by #36

PR #35, opened 2020-07-29
Use click to not interfere with scroll

stephanebisson:
https://phabricator.wikimedia.org/T258263

On mobile, the `touchstart`, `touchmove` and `touchend` events fire for every touch interactions (tap or scroll).

The `click` event represent a single tap and does not fire during a scroll gesture.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -46,7 +46,7 @@ function init( {
 		root.querySelectorAll( selector ),
 		node => {
 			if ( isTouch ) {
-				node.addEventListener( 'touchend', showPopup )
+				node.addEventListener( 'click', showPopup )
 			} else {
 				node.addEventListener( 'mouseenter', showPopup )
 			}
@@ -64,7 +64,7 @@ function init( {
 					if ( isTouch ) {
 						// eslint-disable-next-line no-script-url
 						node.setAttribute( 'href', 'javascript:;' )
-						node.addEventListener( 'touchend', showPopup )
+						node.addEventListener( 'click', showPopup )
 					} else {
 						node.addEventListener( 'mouseenter', showPopup )
 					}


hueitan:
Note :

If we are going to implement this, https://phabricator.wikimedia.org/T257574, touch-enabled desktop (handling both mouse and touch event), we will need to change this `click` event back to `touchend` then handle it in the right way (ex : cancel the event order when detect touchmove)

So we either do it the right solution here or focus only on the mobile experience with `click` for now.

stephanebisson:
> when clicking on the link, it has a gray background on the link for a blink (too fast I can't capture the screen), that's pseudo-class.
> 
> I think you know what I'm saying.

This type of feedback is generally considered useful to indicate that the element was activated but in this case I don't know. I'll ask Sudhanshu.

stephanebisson:
> Note :
> 
> If we are going to implement this, https://phabricator.wikimedia.org/T257574, touch-enabled desktop (handling both mouse and touch event), we will need to change this `click` event back to `touchend` then handle it in the right way (ex : cancel the event order when detect touchmove)
> 
> So we either do it the right solution here or focus only on the mobile experience with `click` for now.

I'm not sure I agree with your assessment here for two reasons:

1. I think `click` is the right way to handle tap on mobile. It has built-in tolerance for non-significant touchmove events. There's almost always some touchmove events. Touch start, move, end, are lower-level events meant to implement gestures.

2. If we do T257574, I think the behavior is going to be fine. Click would try to open the popup on desktop but that would always be a no-op since a mouseenter will happen before the element can be clicked. 

In which case specifically do you see it not working correctly?

hueitan:
> 1. I think `click` is the right way to handle tap on mobile. It has built-in tolerance for non-significant touchmove events. There's almost always some touchmove events. Touch start, move, end, are lower-level events meant to implement gestures.

It's true, I can see the benefit now also for triggering popup in the pinch-zooming behaviour. In addition, it has 300ms delay between tap and click, https://developers.google.com/web/updates/2013/12/300ms-tap-delay-gone-away, but it's longer the common case according to this, so we don't need to worry that.

> 2. If we do T257574, I think the behavior is going to be fine. Click would try to open the popup on desktop but that would always be a no-op since a mouseenter will happen before the element can be clicked.

That makes sense. If we do it, then adding more logic between click and mouseenter that shouldn't be difficult.

hueitan:
> > when clicking on the link, it has a gray background on the link for a blink (too fast I can't capture the screen), that's pseudo-class.
> > I think you know what I'm saying.
> 
> This type of feedback is generally considered useful to indicate that the element was activated but in this case I don't know. I'll ask Sudhanshu.

yeah, this feedback isn't bad actually, similar to the close button of the popup.

PR #36, opened 2020-07-31
Include localized (and optimized) wordmark SVGs

stephanebisson:
https://phabricator.wikimedia.org/T258974

### Problem 
Currently, the wordmark at the top of the preview popup (WikipediA) is static. 

### Solution
Include 22 language-specific variations of the wordmark and select the right one based on the content language.

### Notes
After several rounds of optimization of the SVGs and producing a zip (.gz) version of the production bundle, I was able to bring it down from 406k to 87k, up from ~40k before this change.

I think this is acceptable for now but we'll need to keep an eye on it in the future. Webpack recommends ~244k as the upper limit for a single bundle.


style/preview.less
```diff
@@ -23,17 +24,23 @@
...
+			background-image: inline-svg( '../images/wikipedia-wordmark-en.svg' );
 			background-repeat: no-repeat;
-			background-color: #eaf3ff;
 			background-position: left;
 			background-position-x: 11px;
 			height: 100%;
 			flex-grow: 1;
 			font-size: 24px;
-			padding-left: 13px;
+			margin-left: 13px;
```
hueitan, line 34: background-position-x already have offset 11px, and here gives extra offset, just check on the zeplin, it should have offset left 13px

so either change background position x to 13, and no need for margin left, or the opposite.

style/preview.less
```diff
@@ -23,17 +24,23 @@
...
 			color: #72777d;
 			font-family: 'Linux Libertine', 'Georgia', 'Times', serif;
```
hueitan, line 36: no longer need to font related style - size, color, family, so safe to delete

hueitan:
Note: 

It seems like this is not the full list, for example in pt language, it should have the wikipedia wordmark but no existed here.

style/preview.less
stephanebisson: I did that because the browser I was testing with doesn't support `background-position-x` (FF48) but looking at caniuse, I think this is supported enough that we can use it.

style/preview.less
stephanebisson: right

stephanebisson:
I've pinged design about the inconsistent left padding and the languages missing.

style/preview.less
```diff
@@ -23,30 +24,29 @@
...
-			padding-left: 13px;
-			color: #72777d;
-			font-family: 'Linux Libertine', 'Georgia', 'Times', serif;
+
+			@wordmark-languages: ar, atj, bn, ca, cs, cy, et, fa, fr, he, hi, ja, la, ps, ru, sd, sr, szl, ur, uz, zh-c;
+			each(@wordmark-languages, {
+				&-@{value} {
+					background-image: inline-svg( '../images/wikipedia-wordmark-@{value}.svg' );
+				}
+			});
```
medied, line 39: Neat

stephanebisson:
@hueitan, @medied: ready for review

hueitan:
Looks good, does the tr wordmark get optimized already?

style/preview.less
```diff
@@ -30,14 +30,21 @@
...
+			@wordmark-languages: ar, atj, bn, ca, cs, cy, et, fa, fr, he, hi, ja, la, ps, ru, sd, sr, szl, tr, ur, uz, zh-c;
 			each(@wordmark-languages, {
 				&-@{value} {
 					background-image: inline-svg( '../images/wikipedia-wordmark-@{value}.svg' );
 				}
 			});
-			margin: 0 13px;
+
+			@wordmark-languages-no-margin: ar, bn, fa, he, ps, sd, uz, zh-c;
+			each(@wordmark-languages-no-margin, {
```
hueitan, line 43: I got `CssSyntaxError` when `npm run dev`, but given no more context and I don't see any problem, any one having the same error here?

stephanebisson:
> Looks good, does the tr wordmark get optimized already?

Done

style/preview.less
stephanebisson: I'm not seeing it.

medied: I'm not seeing it either

jpita:
When the modal is loading the content, the wordmark still shows in English before it updates to the target language.
Is this something we can control or the “template” (can’t remember the right word for it) is always in English?
image

stephanebisson:
> When the modal is loading the content, the wordmark still shows in English before it updates to the target language.
> Is this something we can control or the “template” (can’t remember the right word for it) is always in English?
> image

Should be fixed now.

PR #38, opened 2020-08-05
Show popup next to right part of target

stephanebisson:
https://phabricator.wikimedia.org/T256419

### Problem

When the target element is on 2 lines, the popup is positioned incorrectly.

### Solution

find the part of the target the mouse is hovering on and show the popup next to it.


src/popup.js
```diff
@@ -7,22 +7,48 @@ const computePopupPosition = (
...
 	},
 
-	withPx = ( value ) => {
+	withPx = value => {
 		return value ? ( value + 'px' ) : value
 	},
 
+	// Strangely, mouseenter often fires with the pointer slightly
+	// outside any element rect. Making the rects bigger by a few pixel
+	// ensures the pointer will be inside one of them.
```
medied, line 28: Nice

PR #43, opened 2020-08-13
T240007: Image attribution

medied:
https://phabricator.wikimedia.org/T240007

Add image attribution information to the full-screen gallery as requested per design

---

TODO

- [x] Add media info api call
- [x] Display licensing info in fullscreen gallery
- [x] Add "fair use" license
- [x] Sort out caption options
- [x] Fix tests
- [x] Styling refinement
- [x] Explore more unit tests


src/utils.js
```diff
@@ -22,4 +22,38 @@ export const getWikipediaAttrFromUrl = url => {
...
+	},
+
+	defaultParams = {
+		format: 'json',
+		formatversion: 2,
+		origin: '*'
+	},
+
+	buildMwApiUrl = ( lang, params ) => {
+		// TODO - spread or Object.assign?
```
medied, line 34: Spread might have [browser compatibility](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#Browser_compatibility) issues, the linter complains about `Object.assign()`

hueitan:
the image attribution changes after a while without loading animation, if the request takes longer time, the image and info can be wrong visually, do we have the loading animation for image attribution?

IMO, we can keep the current solution, but **hide the image info** while render next image, then **show the info** when the request media info is done.

medied:
>IMO, we can keep the current solution, but hide the image info while render next image, then show the info when the request media info is done.

Yea this could work, let's check in with design

hueitan (commented):
I think this is ready to go? before the final design?

medied:
> I think this is ready to go? before the final design?

Pretty much. I wanted to add one more test to `gallery.test.js` to do a simple check that the attribution divs from `renderImageInfo` get set correctly. This is what it can look like if we use `setTimeout` and `done`:

```
it( 'renders image attribution info', (done) => {
    setTimeout(() => {
        const licenseCC = doc.querySelector( '.wp-gallery-fullscreen-attribution-info-cc' ),
            licenseBY = doc.querySelector( '.wp-gallery-fullscreen-attribution-info-by' ),
            licenseSA = doc.querySelector( '.wp-gallery-fullscreen-attribution-info-sa' ),
            author = doc.querySelector( '.wp-gallery-fullscreen-attribution-info-author' ),
            link = doc.querySelector( '.wp-gallery-fullscreen-attribution-more-info-link' ),
            expectedImageInfo = {
                author: 'Alvesgaspar',
                license: 'CC BY-SA 3.0',
                filePage: 'https://commons.m.wikimedia.org/w/index.php?curid=17960205'
            }
        assert.ok( licenseCC )
        assert.ok( licenseSA )
        assert.ok( licenseBY )
        assert.equal( author.textContent, expectedImageInfo.author )
        assert.equal( link.href, expectedImageInfo.filePage )
        done();
    }, 1500);
} )
```

Because we don't return a Promise... makes it a bit tricky, still trying to figure out if there's a better way. I'm not opposed to merging though, what do you think

src/gallery/fullscreen.js
```diff
@@ -28,8 +29,41 @@ const renderFullScreenGallery = ( lang, dir ) => {
...
+			license = mediaInfo.license,
+			author = mediaInfo.author ? mediaInfo.author : msg( lang, 'gallery-unknown-author' ),
+			link = mediaInfo.filePage,
+			description = getImageDescription()
+
+		return `
+			<div class="wp-gallery-fullscreen-caption">${description}</div>
+			<div class="wp-gallery-fullscreen-attribution">
+				<div class="wp-gallery-fullscreen-attribution-info">
+					${ license && license.indexOf( 'CC' ) !== -1 ? '<div class="wp-gallery-fullscreen-attribution-info-cc"></div>' : ''}
```
stephanebisson, line 58: There must be a cleaner way to treat those 5 cases

style/gallery.less
```diff
@@ -214,6 +217,93 @@
...
+		margin-bottom: 15px;
+
+		&-info {
+			display: flex;
+			align-items: center;
+			margin-left: 16px;
+
+			&-cc {
+				background-image: inline-svg( '../images/attribution-cc.svg' );
+				background-repeat: no-repeat;
```
stephanebisson, line 234: Is it possible to not repeat those 4 attributes for each license?

style/gallery.less
```diff
@@ -214,6 +217,93 @@
...
+				margin-right: 8px;
+			}
+
+			&-author {
+				width: 220px;
+				height: 14px;
+				margin-right: 4px;
+				font-family: 'Lato', 'Linux Libertine', 'Georgia', 'Times', serif;
+				font-size: 12px;
+				font-weight: normal;
```
stephanebisson, line 278: This may have been copy-pasted from zeplin... a whole bunch of `<something>: normal;` is usually not needed.

src/gallery/fullscreen.js
medied: Yea agreed, how about now with `getLicenseInfo`?

medied:
> Maybe we should clear the current attribution while the new one loads. On a slow connection, it is possible to see mismatched image/attribution.

We attempt to do this already with `container.innerText = ''` inside `renderImageInfo` but a couple of thoughts to fine-tune that:

* we could clear the `innerText` before we invoke the API, so take it outside of the callback
* or something along the lines of https://github.com/wikimedia/wikipedia-preview/pull/43#issuecomment-679033749. I believe Sudhanshu is already aware of this idea, it could be better to include this once we tackle final design after slider is complete

stephanebisson:
> * we could clear the `innerText` before we invoke the API, so take it outside of the callback

yes

PR #49, opened 2020-08-19
Storybook

stephanebisson:
https://phabricator.wikimedia.org/T260439

What this PR DOES:
- Add Storybook as a dev dependency
- Show one example of each state (standard, expanded, loading, error, disambiguation)
- Map the components' properties to UI controls so they can be changed
- Add border-radius and drop-shadow to the preview component so it shows correctly here without the need of including popup or touchPopup
- Fix the footer icon style so it stays relative to the footer instead of relative to the larger container. This issue was visible because the icons were shown in the bottom right corner of the screen instead of being inside the preview.
- Export storybook in CI and publish to gh-pages

What this PR does NOT
- Gallery view
- popup and touchPopup components
- bind events

The items about should be considered but I would like to have this PR reviewed and merged without them for a few reasons. First, to avoid potential conflicts with package-lock. Second, working on this made me realize how badly encapsulated some of our components are. Refactoring them is something we should discuss as a team and do together as part of upcoming work rather than all in one go here.


package.json
```diff
@@ -37,6 +43,7 @@
     "mini-css-extract-plugin": "^0.9.0",
     "mocha": "^7.0.1",
     "nyc": "^15.0.0",
+    "react-is": "^16.13.1",
```
hueitan, line 46: i think we don't need this

stephanebisson: this was added by the Storybook installation script. I removed it but it still gets included because it's a dependency of `@storybook/addon-action`.

hueitan (approved):

hueitan:
Q : Should we add storybook link in demo site? Like we said this is the page for developer, so maybe it will interest the developer of our partner?

stephanebisson:
> Q : Should we add storybook link in demo site? Like we said this is the page for developer, so maybe it will interest the developer of our partner?

In the repo description: yes. In the README: probably, yes. But not on the demo site. This is really for a different audience and should not be crowded with links.

.storybook/preview.js
```diff
@@ -0,0 +1,4 @@
+
+export const parameters = {
+  actions: { argTypesRegex: "^on[A-Z].*" },
+}
```
jpita, line 4: We need to fix this? (no new line in end of file)

PR #51, opened 2020-08-20
Add fullscreen gallery story to Storybook

hueitan:
Add to a separate section as gallery can used by *any* 

Annotation 2020-08-20 222549

### DIFF
--- a/src/stories/gallery.stories.js
+++ b/src/stories/gallery.stories.js
@@ -1,0 +1,58 @@
+import { showFullscreenGallery } from '../gallery'
+
+export default {
+	title: 'Gallery',
+	argTypes: {
+		touch: {
+			name: 'Mobile',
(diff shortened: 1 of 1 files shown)


hueitan:
> Good start, would be nice if we can adjust width/height so component looks more like it does from a mobile device screen.
> 
> Also we should think about the merge order we should go about with #43 to include licensing info in this story

agree in both, there is an [addon viewport](https://storybook.js.org/docs/react/essentials/viewport) that able to toggle mobile mode, and let's get #43 merge before this. 👍

hueitan:
Closing this and add this into other PR. 

btw, addon viewport doesn't work well, as it resize after the page load that causes the viewport change so does the gallery width, will do it in other PR or leave it as what it is right now.

PR #54, opened 2020-08-26
T260436: Add unicode support to getWikipediaAttrFromUrl 

medied:
https://phabricator.wikimedia.org/T260436

# Problem

With a link like https://zh.wikipedia.org/wiki/前岐镇, Wikipedia Preview is not enabled via `detectLinks` because the regular expression is unable to extract the article title (`\w` matches latin alphabet words only). 

# Solution 

Add unicode support

```
wikipedia-previews % npm test test/utils.test.js
> wikipedia-preview@1.0.3 test /Users/emedina/internet/wikimedia/wikipedia-previews
> mocha --require ignore-styles,@babel/register,jsdom-global/register "test/utils.test.js"

  getWikipediaAttrFromUrl
    ✓ checks https://en.wikipedia.org/wiki/Cat
    ✓ checks https://en.m.wikipedia.org/wiki/Cat
    ✓ checks https://he.wikipedia.org/wiki/ירח
    ✓ checks https://ru.wikipedia.org/wiki/Каменобродский,_Альфред_Адольфович
    ✓ checks https://zh.wikipedia.org/wiki/太阳帆
    ✓ checks https://ar.wikipedia.org/wiki/كلية_التربية_النوعية_(جامعة_أسوان)
    ✓ checks https://en.wikipedia.org/wiki/Cat#Section
    ✓ checks https://en.wikipedia.org/wiki/Cat?action=edit
    ✓ checks https://de.wikipedia.org/w/index.php?title=Katze
    ✓ checks https://en.m.wikipedia.org/w/index.php?title=Cat&uselang=de
    ✓ checks https://wikimediafoundation.org/

  11 passing (6ms)
```


stephanebisson:
I just pushed another test case to the branch. 

The URLs can always be in on of the two main forms:
/wiki/<title>
/w/index.php?title=<title>

hueitan:
I also add one additional test case for hebrew language. In addition, the unicode includes everything in `\w`, safe to remove?

medied:
Good feedback, how about this update to `regexList`:

* 5 checks in total
* First 3 for `wikipedia.org/wiki/` urls,
    * First two to check whether url has added `#` or `?` values
    * Third option for the basic `wikipedia.org/wiki/`
* Last 2 for the `wikipedia.org/w/index.php`urls
    * One to check for added `&` values
    * Last one for basic `wikipedia.org/w/index.php`
* Use only unicode to simplify

Anything else we're missing?

stephanebisson:
Is there a way to simplify by having a single reg exp per url format that looks for some characters (?, #, &) or nothing?

https://stackoverflow.com/questions/20456467/regex-to-match-string-ending-with-string-or-nothing

medied:
> Is there a way to simplify by having a single reg exp per url format...

How about the approach in https://github.com/wikimedia/wikipedia-preview/pull/54/commits/5bdc41ba4323933c8eb555cac629c05cc3a824a7? Also, I updated [unicode block](https://www.regular-expressions.info/unicode.html) to support latest test (_82.ª_División_Aerotransportada_)

stephanebisson:
Good job making it simpler and so much more robust!

Until we find more problematic cases, I'm going to call this good enough.

PR #55, opened 2020-08-27
POC : render example for gallery row

hueitan:
It solves the same pattern everywhere in each component

**Render util :**
1. Given the container where the elements should render
2. assign the ui element for the class name
3. bind the element event

**Component** (gallery row as the example)
1. Given the template in html string

**There are more to take care, for example** 
 - [ ] Model mode (Ex: Article Summary info in the popup Component)
 - [ ] rerender (replace) when the model change - Virtual Dom change? (Ex: show different article summary)
 - [ ] built-in show/hide/destroy event and allow onShow/onHide/onDestroy event? (Ex: close the fullscreen gallery, hide the popup)
 - [ ] multiple key/action for field `events` (Ex: Chain action?)
 - [ ] Composition for component
 - [ ] etc...

This is just a POC for your reference, and i believe this is a good starting point, idea or use cases are welcomed.

### DIFF
--- a/src/gallery/row.js
+++ b/src/gallery/row.js
@@ -9,7 +9,7 @@ export const GalleryRow = ( mediaItems, lang, dir ) => {
 	return {
 		template: `<div class="wikipediapreview-gallery-row">${images}</div>`,
 		ui: {
-			container: '.wikipediapreview-gallery-image',
+			container: '.wikipediapreview-gallery-row',
 			image: '.wikipediapreview-gallery-image'
 		},
 		events: {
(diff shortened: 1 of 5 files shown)


src/gallery/row.js
```diff
@@ -1,21 +1,22 @@
...
+			image: '.wikipediapreview-gallery-image'
+		},
+		events: {
+			'image click': e => {
+				const selected = e.target.style.backgroundImage.slice( 4, -1 ).replace( /"/g, '' )
+				showFullscreenGallery( mediaItems, selected, lang, dir )
+			}
+		}
+	}
 }
```
hueitan, line 22: The initial idea of this form is given here, https://phabricator.wikimedia.org/P12324, and the Backbone structure.

hueitan:
closing it as we have no plan for this for now.

PR #56, opened 2020-08-28
T260228 Slider

hueitan:
Previous Approach PR : #47 #52 
Slider Component Structure inspired by PR : #55
Phabricator Link : https://phabricator.wikimedia.org/T260228

Steps :
- [x] **(1)** Disable the current image/loading/error code, but don't delete
- [x] **(2)** Create new Slider component / element
- [x] **(3)** Attach all image into slider component
- [x] **(4)** Apply gesture swiping for Slider component
- [x] **(5)** Apply loading animation / error message / refersh button UI and Action to each image (Reuse the current code as much as possible)
    - [x] Loading animation
    - [x] Timeout
    - [x] Image loading error message
    - [x] Refresh button action
- [x] **(6)** Make Sure RTL works
- [x] **(7)** Caption and merge the code with https://github.com/wikimedia/wikipedia-preview/pull/43
    - [x] Current caption (bottom)
    - [x] New caption with image attribution
- [x] Fix Test Specs
- [x] <s>Design tune</s> (The will be another final design ticket for the entire gallery)
    - [x] Image position / center of the screen - take linkedin mobile web page as the example (use 100% instead of vh)
- [x] Add Slider into Storybook (Older PR #51)
- [x] Tested on Desktop Chrome / Mobile Safari Chrome / Edge
- [ ] Tested on IE 11

PR #58, opened 2020-09-02
T259233: Wikipedia Preview card dragging gestures

medied:
https://phabricator.wikimedia.org/T259233

# Problem

Currently, users need to tap 'Continue reading' button expand preview card. We would like to improve the user experience by allowing users to also be able to drag on mobile to interact with preview card

# Solution

Allow users to drag up to expand preview card, drag down to close it, following specified requirements:

- To drag up and reveal more content, user should be able to drag up from either the body or header of the card
- To drag down and close the Wikipedia Preview card, users should only be able to drag down the card header. This is to not interfere with any possible text scrolling
- The tap gesture should remain unchanged

Solution inspired from #56 

---

#### TODOs

- [x] Design clarification: swipe up initial touch position
- [x] Design clarification: swipe down gesture when scrolling text is present

### DIFF
--- a/src/event.js
+++ b/src/event.js
@@ -94,7 +94,7 @@ export const customEvents = popup => {
 				handleTouchEnd = ( isHeader ) => {
 					const expanded = element.querySelector( '.wikipediapreview.expanded' ),
 						delta = initialY - finalY,
-						isOverThreshold = Math.abs( delta ) > 60,
+						isOverThreshold = Math.abs( delta ) > 80,
 						isNotExpandedBody = !expanded && !isHeader || isHeader
 
 					if ( delta < 0 && isOverThreshold && isNotExpandedBody ) {
--- a/src/event.js
+++ b/src/event.js
@@ -67,15 +67,15 @@ export const customEvents = popup => {
 		applyDragEvent = ( element ) => {
 			let initialY,
 				finalY,
-				containerBodyStyle,
+				previewBodyStyle,
 				initialHeight
 
-			const containerHeader = element.querySelector( '.wikipediapreview-header' ),
-				containerBody = element.querySelector( '.wikipediapreview-body' ),
+			const previewHeader = element.querySelector( '.wikipediapreview-header' ),
+				previewBody = element.querySelector( '.wikipediapreview-body' ),
 				handleTouchStart = ( e ) => {
 					initialY = e.touches[ 0 ].clientY
-					containerBodyStyle = window.getComputedStyle( containerBody )
-					initialHeight = Number( containerBodyStyle.height.slice( 0, -2 ) )
+					previewBodyStyle = window.getComputedStyle( previewBody )
+					initialHeight = Number( previewBodyStyle.height.slice( 0, -2 ) )
 				},
 
 				handleTouchMove = ( e, isHeader ) => {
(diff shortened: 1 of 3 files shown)


hueitan (changes requested):
Something missing and recommend to do it

1. Before expand, drag down should close the popup
1. After expand, does drag down more than 60px too less? too sensitive?
1. once drag match the threshold, apply the animation drag to the expanded (closed) position
1. Tap CTA (Continue reading) doesn't expand the card height
1. "throw" event that require duration calculation.

src/event.js
```diff
@@ -64,6 +64,61 @@ export const customEvents = popup => {
...
+					const clientY = e.touches[ 0 ].clientY,
+						offset = initialY - clientY,
+						currentHeight = initialHeight + offset,
+						expanded = element.querySelector( '.wikipediapreview.expanded' )
+
+					finalY = clientY
+					if ( !isHeader && !expanded || isHeader && expanded ) {
+						containerBody.style.maxHeight = currentHeight + 'px'
+					}
+				},
```
hueitan, line 92: insert the code `e.preventDefault()` here to prevent lagging sometimes

medied: Looks like adding `e.preventDefault()` messes with the text scrolling actually

medied:
>    Before expand, drag down should close the popup
>    After expand, does drag down more than 60px too less? too sensitive?
>    Tap CTA (Continue reading) doesn't expand the card height

Cool, addressed 1, 2 & 4

>   once drag match the threshold, apply the animation drag to the expanded (closed) position
 >  "throw" event that require duration calculation.

Could you please elaborate on 3 & 5?

hueitan:
> Could you please elaborate on 3 & 5

> 3. once drag match the threshold, apply the animation drag to the expanded (closed) position

Currently, when the `touchend` event apply, the preview card is instantly expanded or hide, it's better to have the animation (transition in css) instead of sudden appear sudden hide

5. "throw" event that require duration calculation.

quick swipe within the given threshold, similar to gallery slider, quick swipe to get the next image instead of swiping more than 40% of the image, this gives a better experience for user who knows how to swipe.

src/event.js
hueitan: Opsss, then better don't have this code.

medied:
> > Could you please elaborate on 3 & 5
> 
> > 1. once drag match the threshold, apply the animation drag to the expanded (closed) position
> 
> Currently, when the `touchend` event apply, the preview card is instantly expanded or hide, it's better to have the animation (transition in css) instead of sudden appear sudden hide
> 
> > 1. "throw" event that require duration calculation.
> 
> quick swipe within the given threshold, similar to gallery slider, quick swipe to get the next image instead of swiping more than 40% of the image, this gives a better experience for user who knows how to swipe.

Alright I'd like to share this exploration sub-branch and have at least you and Sudhanshu play with it and see how it feels:

`T259233-swipe-more-content-animation-exploration`: https://github.com/wikimedia/wikipedia-preview/commit/24d75c208afe815bbf6ca7642c8011ee13416b0d

Some observations & thoughts: 

* The transition on dragging up feels about right to me. There's transition when the card is dragged within the threshold as well. 
* What could need some more refinement is the transition when dragging down, past the threshold to close the card
  * Since we actually switch the `visibility` to `hidden`, it still may feel a bit sudden
  * Doing some research and looking around at some solutions, what may feel better is to apply a slight "reverse animation" where the card height reduces to some point and then `visibility` gets set to `hidden`
  * But I want you all to play with it first for feedback, it may be the case as well that it feels fine the way it is right now in `T259233-swipe-more-content-animation-exploration`
* Given the current threshold, it's possible the "quick swipe/drag" is not necessary. Would like to get Sudhanshu's input on that as well.

hueitan (approved):
Nice gesture!

PR #61, opened 2020-09-14
T257648 Desktop expand preview popup layout (max height calculation)

hueitan:
Phabricator Ticket : https://phabricator.wikimedia.org/T257648

Problem: 

The preview popup may overflow behind the address bar when it expands.

Solution:

Fix the preview popup fix under the address bar when it exceeds the maximum height for the preview body

tested on

- [x] chrome
- [x] firefox
- [x] IE 11
- [x] Edge

### DIFF
--- a/style/preview.less
+++ b/style/preview.less
@@ -309,8 +309,9 @@
 
 	&.expanded {
 		.wikipediapreview-body {
-			max-height: 496px;
-			overflow: hidden;
+			// calculate the max height in onExpand()
+			// max-height: 496px;
+			overflow: scroll;
 
 			&:after {
 				content: none;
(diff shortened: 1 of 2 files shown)


stephanebisson (approved):
LGTM

style/preview.less
```diff
@@ -309,8 +309,9 @@
 
 	&.expanded {
 		.wikipediapreview-body {
-			max-height: 496px;
-			overflow: hidden;
+			// calculate the max height in onExpand()
+			// max-height: 496px;
```
medied, line 313: Did you mean to leave this `max-height` commented out?

medied (approved):

style/preview.less
hueitan: yes, I can remove that, but it gives a highlight where the height of the expanded came from which is useful for debug.

medied: Yea figured you were thinking something along that line, I'm fine with leaving it like that

PR #63, opened 2020-09-16
T261745: gallery with best fit image

medied:
https://phabricator.wikimedia.org/T261745

# Problem

There are currently a few places where the gallery could be improved:

- it loads very big and heavy images for the fullscreen view
- it's missing the white background for transparent images
- has this issue: [T259854](https://phabricator.wikimedia.org/T259854)
- does not take advantage of the full screen for image layout

# Solution

This ticket is aims to optimize the gallery with respect to some these. Specifically:
        
- use of `iiurlwidth` and `iiurlheight` props to make each fullscreen image as light as possible while still looking good (non-pixelated) on device
  - see the table below for different size comparisons 
- include white background for transparent images
- resolves T259854
- use maximum screen size for image with the image info fixed on top with the same gallery background black color, as a temporary solution.

---

This table shows the difference in sizes, but the final positioning of the image is different after b122a2a977c1c841f4f63a070cc066a95f2375aa

heavier image | lighter image
-- | --
image_2550 x 1480_  | image  _375 x 217_
image _6000 x 4000_ | image _375 x 217_
image _2832 x 3730_ | image _375 x 494_

---

TODO 
- [x] update `requestPageMediaInfo` tests

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -1,5 +1,5 @@
 import { cachedRequest } from './cachedRequest'
-import { buildMwApiUrl, buildCommonsApiUrl, convertUrlToMobile, strip } from './utils'
+import { buildMwApiUrl, buildCommonsApiUrl, convertUrlToMobile, strip, getDeviceSize } from './utils'
 
 const requestPagePreview = ( lang, title, isTouch, callback, request = cachedRequest ) => {
 		const url = `https://${lang}.wikipedia.org/api/rest_v1/page/summary/${encodeURIComponent( title )}`
@@ -54,6 +54,8 @@ const requestPagePreview = ( lang, title, isTouch, callback, request = cachedReq
 				iiextmetadatafilter: 'License|LicenseShortName|ImageDescription|Artist',
 				iiextmetadatalanguage: lang,
 				iiextmetadatamultilang: 1,
+				iiurlwidth: getDeviceSize().width,
+				iiurlheight: getDeviceSize().height,
 				iiprop: 'url|extmetadata',
 				titles: title
 			},
@@ -74,13 +76,15 @@ const requestPagePreview = ( lang, title, isTouch, callback, request = cachedReq
 					( typeof ImageDescription.value === 'string' && ImageDescription.value ) ||
 					( ImageDescription.value[ lang ] ||
 						ImageDescription.value[ Object.keys( ImageDescription.value )[ 0 ] ] )
-				)
+				),
+				imageUrl = imageInfo[ 0 ].thumburl
 
 			return {
 				author,
 				description,
 				license: LicenseShortName && LicenseShortName.value,
-				filePage: convertUrlToMobile( imageInfo[ 0 ].descriptionshorturl )
+				filePage: convertUrlToMobile( imageInfo[ 0 ].descriptionshorturl ),
+				bestFitImageUrl: imageUrl
 			}
 		}, callback )
 	}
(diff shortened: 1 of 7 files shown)


hueitan:
This is exactly the solution that in my mind, great job!

style/gallery.less
```diff
@@ -84,13 +84,14 @@
...
-				visibility: hidden;
+				background-color: #fff;
 			}
 
 			&-caption {
 				position: absolute;
+				background-color: #222;
 				bottom: 35px;
-				width: calc( 100vw - 32px );
+				width: calc( 100vw + 1px );
```
hueitan, line 94: _very strange, right? but I found this magic number here helps in the ios safari browser._ what's your finding here?

test/api.test.js
```diff
@@ -239,6 +235,9 @@ describe( 'requestPageMediaInfo', () => {
 						imagerepository: 'local',
 						imageinfo: [
 							{
+								thumburl:	'https://upload.wikimedia.org/wikipedia/commons/thumb/0/08/Horn_Louvre_OA4069.jpg/375px-Horn_Louvre_OA4069.jpg',
```
hueitan, line 238: remember to run `npm run test:fix-lint`

src/api.js
```diff
@@ -54,6 +52,8 @@ const requestPagePreview = ( lang, title, isTouch, callback, request = cachedReq
 				iiextmetadatafilter: 'License|LicenseShortName|ImageDescription|Artist',
 				iiextmetadatalanguage: lang,
 				iiextmetadatamultilang: 1,
+				iiurlwidth: getDeviceSize().width,
+				iiurlheight: getDeviceSize().height,
```
hueitan, line 56: instead of having full height, the exact height of the image shown on the screen should be deducted from the minimum image attribution height.

hueitan (changes requested):

style/gallery.less
medied: Duude... yes, very strange indeed. Took me a while to realize that was breaking the sliding for the caption, as in everything was sliding except the caption (which would remain the same from the first image rendered in full screen). This might change later when we work on https://phabricator.wikimedia.org/T262373. For the image info (caption/licensing) to be fixed on top of the image I needed the width to cover 100vw of the width, but it wouldn't work without `calc`

src/api.js
medied: Could you elaborate on this point?

style/gallery.less
hueitan: Yes, we can check this again on another ticket.

src/api.js
hueitan: Screenshot 2020-09-18 113822

the red square size is the real width/height for the image

Does this screenshot help to explain?

medied: Ohh so you're saying instead of requesting `iiurlheight: getDeviceSize().height` to the api, we do something like `iiurlheight: getDeviceSize().height - 100`? That is, we substract the minimum image attribution height, which is around ~100 px.

Requesting `iiurlheight: getDeviceSize().height - 100` gives us the following

Screen Shot 2020-09-18 at 9 37 18 AM

So we could do that... but I understood [Sudhanshu's point](https://phabricator.wikimedia.org/T261745#6466762) on using "maximum screen size" as in using full height.

hueitan: you are right, "maximum screen size", let's keep it 100% and don't minus it.

hueitan (approved):
Looks Good

jpita:
THE SPEEEEEED!!! THE SPEEEEEED!!!! 😍😍😍😍😍😍😍😍😍😍

PR #65, opened 2020-09-21
T257648 follow up for desktop layout

hueitan:
Phabricator Ticket : https://phabricator.wikimedia.org/T257648
Previous PR : #61 

Follow up list :

 - [x] Fix for IE 11
 - [x] Disable Scroll bar
 - [x] Apply the same logic when opening popup preview, https://phabricator.wikimedia.org/T257648#6470315


stephanebisson (changes requested):
Currently, `popup.js` has the logic and the math to position the popup correctly relative to the target word and screen size. It would be great for this new code to also live in that file to try to keep the logic together. `event.js` can have the trigger but delegate the work like `popup.expand(...)` or something.

hueitan:
> Currently, `popup.js` has the logic and the math to position the popup correctly relative to the target word and screen size. It would be great for this new code to also live in that file to try to keep the logic together. `event.js` can have the trigger but delegate the work like `popup.expand(...)` or something.

I like to do this but something comes into my mind.

1. The `popup.js` math position is calculating the position of the popup, and we assume it doesn't know anything related to the preview content (preview template).
1. This logic checks the preview content height pixels, it's within the preview scope; Plus, the `popup.js` has no access to expand logic, that belongs to the preview content logic.

Therefore the `expand height calculation` logic stays within `event.js`, similar to all expand related logic.

Any thoughts? 💭

stephanebisson (approved):

PR #66, opened 2020-09-22
T263544 Shows gallery row when opening up the preview popup if there is enough space

hueitan:
Phab ticket : https://phabricator.wikimedia.org/T263544

Problem
===

Go to the desktop site, https://wikimedia.github.io/wikipedia-preview/demo/articles/hindi.html, hover the second link `हुमांयू`, it shows an empty gallery row

Solution
===

The value should be set before showing the preview

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -35,14 +35,15 @@ function init( {
 				if ( popup.loading ) {
 					if ( data ) {
 						if ( data.type === 'standard' ) {
+							popup.lang = lang
+							popup.title = title
+
 							popup.show(
 								renderPreview( lang, data, isTouch ),
 								target,
 								pointerPosition
 							)
 
-							popup.lang = lang
-							popup.title = title
 							const expanded = root.querySelector( '.wikipediapreview.expanded.mobile' )
 
 							if ( expanded && popup.expand ) {
--- a/src/index.js
+++ b/src/index.js
@@ -43,12 +43,6 @@ function init( {
 								target,
 								pointerPosition
 							)
-
-							const expanded = root.querySelector( '.wikipediapreview.expanded.mobile' )
-
-							if ( expanded && popup.expand ) {
-								popup.expand()
-							}
 						} else if ( data.type === 'disambiguation' ) {
 							popup.show(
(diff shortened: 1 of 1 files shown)


hueitan:
oh wait, this fix brings the duplicate image row

src/index.js
```diff
@@ -35,19 +35,14 @@ function init( {
...
 								renderPreview( lang, data, isTouch ),
 								target,
 								pointerPosition
 							)
-
-							popup.lang = lang
-							popup.title = title
-							const expanded = root.querySelector( '.wikipediapreview.expanded.mobile' )
-
-							if ( expanded && popup.expand ) {
```
hueitan, line 48: @stephanebisson @medied do you remember why we need this here?

in `event.js`'s `onShow` event, we already have the function called `onExpand()`

stephanebisson: I don't remember... it seems redundant. It's OK to remove but we have to test the hell out of it.

medied: I think it's for the `requestPageMedia` and subsequent `getGalleryRow` for when the preview text is short enough to render with gallery row... but yea this can probably be improved

image

hueitan: > I think it's for the `requestPageMedia` and subsequent `getGalleryRow` for when the preview text is short enough to render with gallery row... but yea this can probably be improved

Got it, and there is already a check during `onShow` event, that should be enough.

hueitan:
ready to review, and I think it's worth @jpita for a quick check

Note : This has no relevant phabricator ticket, it's a bug within the codebase.

stephanebisson:
> Note : This has no relevant phabricator ticket, it's a bug within the codebase, or should I create one?

It's always best to have a task with clear steps to reproduce the issue.

jpita:
I saw this for a moment and now I can't replicate it 😂 
image

hueitan:
> I saw this for a moment and now I can't replicate it 😂

The second commit of this PR is to fix this, so you won't get this 😃

src/index.js
hueitan: oh wait, i can see the reason why now @medied, let me rework on this.

hueitan:
close and re open it later.

PR #67, opened 2020-09-22
T260408: preview wordmark refinement

medied:
https://phabricator.wikimedia.org/T260408

# Problem 

Currently, the Wikipedia wordmark on preview has a pale blue background which needs to be updated to match our visual identity guidelines.

# Solution

Update preview header according to design spec

Before  | After
----- | -----
Screen Shot 2020-09-22 at 13 39 28 | Screen Shot 2020-09-22 at 13 39 13


hueitan (changes requested):
this is the only thing I can found

without thumbnail case, the wordmark and x should fall on the same line.
image

hueitan (approved):
LGTM

src/preview.js
```diff
@@ -76,19 +75,15 @@ const renderPreview = ( lang, data, isTouch ) => {
...
+					${getPreviewHeader( lang )}
 					<div class="wikipediapreview-disambiguation">
 						<div class="wikipediapreview-disambiguation-body">
 							<div class="wikipediapreview-disambiguation-body-message">
 								<div class="wikipediapreview-disambiguation-body-icon"></div>
 								<span>${msg( lang, 'preview-disambiguation-message', title )}</span>
 							</div>
 							<div class="wikipediapreview-disambiguation-body-readon">
-								<a href=${`https://${lang}.m.wikipedia.org/wiki/${encodeURIComponent( title )}`} target="_blank">${msg( lang, 'read-on-wiki' )}</a>
+								<a href=${`https://${lang}.m.wikipedia.org/wiki/${encodeURIComponent( title )}`} target="_blank">${msg( lang, 'read-on-wiki-disambiguation' )}</a>
```
stephanebisson, line 86: We decided not to introduce a distinct CTA here when we did the disambiguation task. I'm guessing the mockups were never updated so you saw the discrepancy.

stephanebisson (changes requested):
Please remove the disambiguation CTA

stephanebisson (approved):

PR #70, opened 2020-09-23
T256308: offline mode

medied:
https://phabricator.wikimedia.org/T256308

# Summary

Adds offline mode according to design spec:

> The offline state shows up when the user's device is not connected to the internet.
    Tapping on Try Again will show the loading state and if still no connection available then control comes back to offline state.
    Users can close the preview by tapping on a close icon or area outside.
    Present media viewer error state if a user loses internet connection while viewing the image. Refresh CTA will show media viewer loading state(bouncing dots) and if still no connection available then control comes back to offline state.


src/utils.js
```diff
@@ -21,6 +21,10 @@ export const getWikipediaAttrFromUrl = url => {
 	isTouch = 'ontouchstart' in window || ( navigator.maxTouchPoints > 0 ) ||
 		( navigator.msMaxTouchPoints > 0 ),
 
+	isOnline = () => {
```
stephanebisson, line 24: With a lambda you can make this a one-liner:

```
isOnline = () => window.navigator.onLine
```

style/preview.less
```diff
@@ -286,6 +286,51 @@
...
+			&-message {
+				display: flex;
+				margin-left: 16px;
+				margin-right: 16px;
+				font-size: 16px;
+				font-weight: normal;
+				font-stretch: normal;
+				font-style: normal;
+				line-height: 1.4;
+				letter-spacing: normal;
```
stephanebisson, line 307: All those `<something>: normal` may not be needed.

Same with the &-retry block below.

src/event.js
```diff
@@ -36,7 +36,8 @@ export const customEvents = popup => {
 		getPreviewBody = ( element ) => {
 			return element.querySelector( '.wikipediapreview-body' ) ||
 				element.querySelector( '.wikipediapreview-disambiguation' ) ||
-				element.querySelector( '.wikipediapreview-error' )
+				element.querySelector( '.wikipediapreview-error' ) ||
+				element.querySelector( '.wikipediapreview-offline' )
```
stephanebisson, line 40: I really dislike this code. It's not you fault, and you're doing the right thing by adding the offline body class here but really, all states should just have the same `.wikipediapreview-body` element that can contain various things.

stephanebisson (changes requested):
Please add the offline state to the storybook

src/gallery/slider.js
```diff
@@ -99,13 +100,16 @@ const clientWidth = window.innerWidth,
 			errorElement = container.querySelector( `.${prefixClassname}-item-loading-error` )
 
 		if ( refresh ) {
-			imageElement.src = `${imageElement.src}?timestamp=${Date.now()}`
```
hueitan, line 102: the timestamp is needed here because it tells the browser to retry the image, if no timestamp was given the browser will treat it like before state (possibly in cache or others)

src/event.js
medied: Yea agree they should all have the same `wikipediapreview-body`, that will make the code more robust. This was evident when I added the offine preview card and saw it had broken `applyDragEvent`. Will take a look at how we can improve this.

src/gallery/slider.js
medied: Hmm ok, how about including the `?timestamp` within `showImageInfo` when setting the image `src` there?

https://github.com/wikimedia/wikipedia-preview/pull/70/files#diff-d9d613b93342c10d9ee8b8ac7c9118f6R158

I updated this logic here to use `showImageInfo` instead of directly resetting the `src` because we need the api call to go through again when user is coming back from offline

hueitan: > I updated this logic here to use showImageInfo instead of directly resetting the src because we need the api call to go through again when user is coming back from offline

Make sense

> how about including the ?timestamp within showImageInfo when setting the image src there

It should work, but that means the image never **gets cached by the browser**, reload the image every time when the user comes back to the gallery slider view again.

Probably another parameters `showImageAndInfo( index, refreshImage = false)` and add timestamp when `refreshImage ` is true?

src/event.js
medied: Updated with https://github.com/wikimedia/wikipedia-preview/pull/70/commits/21e24468bd76fece030d3037de97df86c6c43e28

stephanebisson:
Looks good. 

Minor tweak needed to the message and icon alignment in RTL.

LTR

RTL

stephanebisson (dismissed):

stephanebisson:
> Any of these issues is to be fixed on this branch?

I think all of them should be fixed here, or at least investigated.

medied:
Thanks @jpita, please feel free to try again. My last update (https://github.com/wikimedia/wikipedia-preview/pull/70/commits/d2998dc148767c64551241040dd62d05d7af74b0) should address all of your points. I am opting for triggering the refresh process from `renderNext` instead of `showImageAndInfo` for a more comprehensive gallery refresh when network is back on.

Also including https://github.com/wikimedia/wikipedia-preview/pull/70/commits/be7a2cd6b422cc311c4f4e3bf0dcf442aea1cbdd with regards to the desktop issue.

src/gallery/slider.js
```diff
@@ -99,13 +100,26 @@ const clientWidth = window.innerWidth,
...
+			const slider = parentContainer.querySelector( `.${prefixClassname}` ),
+				items = slider.querySelectorAll( `.${prefixClassname}-item` )
+
+			items.forEach( item => {
+				const image = item.querySelector( 'img' )
+				if ( image ) {
+					item.removeChild( image )
+				}
+			} )
```
hueitan, line 111: the event and refresh are for the individual image refresh purpose, why are we remove all the images within the slider?

medied: Right we were originally refreshing each individual image and that's why we were seeing the behavior that [Pita described](https://github.com/wikimedia/wikipedia-preview/pull/70#issuecomment-700316653). The idea here now is to accomplish a more comprehensive refresh by triggering a `renderNext` which consequently triggers three `showImageAndInfo` calls with different offsets (-1, 0, 1). For the `renderNext` --> `showImageAndInfo` cycle to complete successfully, we have to remove the `img` elements that are being replaced (supposedly failed images becuase of the connection issues). I'm opting for removing all of the `img` elements just to be on the safe side and ensure that there will be no `img` elements conflicts. If the network is back on, all images would be successfully loaded again. Does that make sense?

hueitan: alright, this can make sure it loads all the images so user doesn't have to refresh each image repeatedly, which is not bad.

hueitan (approved):
@jpita ready for test again

jpita:
> if after step 9 we close the gallery and bring it back up, the "refreshed" image is no longer showing, only shows the first 2


@medied this issue is still happening.
I load 4 images then cut the internet.
Once I close the gallery and then bring it back up , only 2 images are available

medied:
>I load 4 images then cut the internet.
Once I close the gallery and then bring it back up , only 2 images are available

I'm trying to replicate this behavior from my end and I'm seeing slightly different results depending on browser. For example, when I test in Chrome I'm seeing all images display again on opening the gallery a second time while still offline. If I try the same in Firefox, only some images display while others (that had loaded on first opening) show up with offline messaging. 

So it's interesting, this leads me to believe that the cache system for each browser is probably handling things differently and hence the different results. If we look at the method that gets executed when users close the fullscreen gallery, we are actually removing the fullscreen gallery entirely:

```
	hideFullscreenGallery = container => {
		const fullscreenGallery = container.querySelector( `.${prefixClassname}` )
		container.removeChild( fullscreenGallery )
	}
```

@hueitan @stephanebisson any thoughts? I'm leaning towards thinking that the UX and functionality is okay as is, but maybe I'm missing something we could do to improve it.

medied:
@jpita could you please confirm if you are seeing different behaviors depending on browser like I am? Or otherwise describe what behavior you see for different browsers.

@hueitan I am assuming you also see a cache difference for each browser, per your comment.

If we all agree on a cache difference per browser I believe that will help us make a decision on how to move forward here.

jpita:
Just noticed a new behaviour on iPhone:
Site reloads and closes the gallery when the internet is reactivated when I tap on "Refresh" or swipe to a different image.

This is happening when I'm testing with the app running on my laptop and I connect to it through IP.
Not sure why this is happening.
@medied please schedule a meeting with me when you can during your morning (this Thursday or next week) so we can try to debug what's happening (iPhone is a bitch to debug).

medied:
>Site reloads and closes the gallery when the internet is reactivated when I tap on "Refresh" or swipe to a different image. This is happening when I'm testing with the app running on my laptop and I connect to it through IP.

Working with Pita on this, behavior seems to be related to the localhost/IP set up we use to test on device. Conside this 'reversed' test, so to speak: if we test what's on `master` both on the demo site and via the localhost/IP set up, I'm seeing the same behavior: it auto-reloads in the localhost/IP setup, but it doesn't auto-reload on the demo site once connection is re-established

PR #73, opened 2020-09-28
Fix wordmark position in RTL and add to Storybook

stephanebisson:
Fix the margin around the wordmark with and without thumbnail. Also add those cases to the Storybook.


medied (commented):
LGTM

medied (approved):

PR #75, opened 2020-09-29
Use system font

stephanebisson:
https://phabricator.wikimedia.org/T260407

Use system fonts according to the design style guide.

Also removed a bunch of `<something>:normal` useless CSS directives.


medied (approved):
Looks good!

PR #77, opened 2020-10-05
Load wordmark SVGs

stephanebisson:
https://phabricator.wikimedia.org/T262289

Wordmark SVGs make the bundle too heavy. 

Load them from wikipedia.

Also introduces webpack-bundle-analyzer with `npm run wba`


hueitan (approved):
Nice one, I like this.

PR #79, opened 2020-10-15
update link style in rtl direction

hueitan:
slightly different but there is

| Before | After |
| --- | --- |
| localhost_8080_articles_arabic html(iPhone 6_7_8) (1) | localhost_8080_articles_arabic html(iPhone 6_7_8) |


medied (approved):
LGTM

PR #80, opened 2020-10-20
T263733 Shows the latest hover link (Race condition edge case)

hueitan:
Phab ticket : https://phabricator.wikimedia.org/T263733

Problem 
===

Once hover action is called, the callback event will execute no matter which hovers user have

Solution
===

Remember the user hover id, and do the callback only the hover id and the current hover id is matched

Note
===

We don't clear the request event, that can be reserved as the cache for the next time the user hover it.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -5,6 +5,8 @@ import { createTouchPopup } from './touchPopup'
 import { renderPreview, renderLoading, renderError, renderDisambiguation, renderOffline } from './preview'
 import { getWikipediaAttrFromUrl, isTouch, getDir, isOnline } from './utils'
 
+let currentPopupId
+
 function init( {
 	root = document,
 	selector = '[data-wikipedia-preview]',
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -17,7 +19,8 @@ function init( {
 			createPopup( popupContainer ),
 		events = customEvents( popup ),
 		last = {},
-		showPopup = ( e, refresh = false ) => {
+		showPopup = ( e, refresh = false, popupId = Date.now() ) => {
```
stephanebisson, line 22: Is `showPopup()` ever called with a `popupId`?

hueitan: nope, i'm thinking of where's the best place to set it.

stephanebisson: Can't you do `currentPopupId = Date.now()` directly?

hueitan: Done

stephanebisson (approved):

medied (approved):
Very neat

PR #81, opened 2020-10-20
T260364: Reduce code duplication for preview states

medied:
https://phabricator.wikimedia.org/T260364

# Problem
> The file preview.js contains the rendering code for the 5 states of the preview: standard, loading, error, disambiguation, offline. They have almost identical chrome and layout but the markup and style are duplicated.

# Solution

Refactoring with https://github.com/wikimedia/wikipedia-preview/commit/809c6d4f55eb6325e2ba6ab8437a67faf4f26281:
* Net difference in preview.js is even (_33 additions & 33 deletions_) but it should be easier to maintain this way
* Reduction of code lines come through preview.less. I opted to retain whatever specific CSS for each state as they were.
* The [storybook](https://github.com/storybookjs/storybook) was really useful to develop and test different states. Also tested on device.

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -5,7 +5,7 @@
     "gallery-loading-error-refresh": "Refresh",
     "gallery-loading-still": "Still loading",
     "gallery-unknown-author": "Author unknown",
-    "preview-loading-error": "There was an issue while displaying this preview.",
+    "preview-error-message": "There was an issue while displaying this preview.",
     "read-on-wiki": "Read on Wikipedia",
     "read-more": "Read more on Wikipedia",
     "preview-disambiguation-message": "Title <strong>$1</strong> is related to more than one article on Wikipedia.",
--- a/src/preview.js
+++ b/src/preview.js
@@ -12,41 +12,34 @@ const getPreviewHeader = ( lang, imageUrl = '' ) =>{
 	`.trim()
 	},
 
-	getPreviewBody = ( lang, title, type, isTouch = false ) => {
-		const getMessage = () => {
-				if ( type === 'disambiguation' ) {
-					return `<span>${msg( lang, `preview-${type}-message`, title )}</span>`
-				}
-				return `<span>${msg( lang, `preview-${type}-message` )}</span>`
-			},
-
-			getCallToAction = () => {
-				if ( type === 'offline' ) {
-					return `<a>${msg( lang, 'preview-offline-cta' )}</a>`
-				}
-				return `<a href="${buildWikipediaUrl( lang, title, isTouch )}" target="_blank">${msg( lang, 'read-on-wiki' )}</a>`
-			}
-
+	getPreviewBody = ( type, message, cta ) => {
 		return `
 			<div class="wikipediapreview-body">
-				<div class="wikipediapreview-body-template-${type}">
+				<div class="wikipediapreview-body-template ${type}">
 					<div class="wikipediapreview-body-template-message">
 						<div class="wikipediapreview-body-template-icon"></div>
-							${getMessage()}
+							${message}
 					</div>
 					<div class="wikipediapreview-body-template-action">
-						${getCallToAction()}
+						${cta}
 					</div>
 				</div>
 			</div>
 	`.trim()
 	},
 
-	renderPreview = ( lang, data, isTouch ) => {
-		const imageUrl = data.imgUrl
+	render = ( lang, isTouch, dir, headerContent, bodyContent ) => {
 		return `
-			<div class="wikipediapreview${isTouch ? ' mobile' : ''}" lang="${lang}" dir="${data.dir}">
-				${getPreviewHeader( lang, imageUrl )}
+			<div class="wikipediapreview ${isTouch ? 'mobile' : ''}" lang="${lang}" dir="${dir}">
+				${headerContent}
+				${bodyContent}
+			</div>
+		`.trim()
+	},
+
+	renderPreview = ( lang, data, isTouch ) => {
+		const imageUrl = data.imgUrl,
+			bodyContent = `
 				<div class="wikipediapreview-body">
 					${data.extractHtml}
 					<div class="wikipediapreview-gallery">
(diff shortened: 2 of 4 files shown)


src/preview.js
```diff
@@ -12,6 +12,36 @@ const getPreviewHeader = ( lang, imageUrl = '' ) =>{
 	`.trim()
 	},
 
+	getPreviewBody = ( lang, title, type, isTouch = false ) => {
+		const getMessage = () => {
+				if ( type === 'disambiguation' ) {
```
stephanebisson, line 17: Since we have distinct entry points (renderPreview, renderError, etc) we shouldn't have those conditionals. Each entry point should compose it's own UI based on re-usable elements. In other words, `renderDisambiguation()` is the only one to know or care about its own body content.

src/preview.js
```diff
@@ -57,17 +87,7 @@ const getPreviewHeader = ( lang, imageUrl = '' ) =>{
 		return `
 			<div class="wikipediapreview expanded ${isTouch ? 'mobile' : ''}" lang="${lang}" dir="${dir}">
```
stephanebisson, line 88: This line is still duplicated in all entry points. It could be extracted into some kind of `render( lang, isTouch, headerContent, bodyContent, ctaContent )` that handles the general container for all states.

style/preview.less
```diff
@@ -93,6 +130,35 @@
 			font-size: 18px;
 			padding: 10px 20px;
 		}
+
+		&-template-error {
+			&:extend( .wikipediapreview-body-template );
```
stephanebisson, line 135: The use of `extend` in dev but doesn't reduce the bundle size since it inlines the content during compilation. 

CSS is already pretty good at having something like a `body` class with all the common directives and a bunch of `body-error`, `body-disambiguation`, etc with overrides.

i18n/en.json
```diff
@@ -5,7 +5,7 @@
     "gallery-loading-error-refresh": "Refresh",
     "gallery-loading-still": "Still loading",
     "gallery-unknown-author": "Author unknown",
-    "preview-loading-error": "There was an issue while displaying this preview.",
+    "preview-error-message": "There was an issue while displaying this preview.",
```
stephanebisson, line 8: You also need to rename the message in the qqq file.

stephanebisson (changes requested):
Very good first pass. This is a well needed cleanup.

I think we can and should go further, see inline comments.

i18n/en.json
medied: > Very good first pass. This is a well needed cleanup.
> 
> I think we can and should go further, see inline comments.

Great ideas from inline comments, thanks for the on point feedback. My interpretation of it now reflected in https://github.com/wikimedia/wikipedia-preview/pull/81/commits/0e014a1c411a9d3a5b6e37b1091388588db8f643, what do you think?

medied:
> Very good first pass. This is a well needed cleanup.
> 
> I think we can and should go further, see inline comments.

Great ideas from inline comments, thanks for the on point feedback. My interpretation of it now reflected in https://github.com/wikimedia/wikipedia-preview/pull/81/commits/0e014a1c411a9d3a5b6e37b1091388588db8f643, what do you think?

src/preview.js
```diff
@@ -12,11 +12,34 @@ const getPreviewHeader = ( lang, imageUrl = '' ) =>{
 	`.trim()
 	},
 
-	renderPreview = ( lang, data, isTouch ) => {
-		const imageUrl = data.imgUrl
+	getPreviewBody = ( type, message, cta ) => {
+		return `
+			<div class="wikipediapreview-body">
+				<div class="wikipediapreview-body-template ${type}">
```
hueitan, line 18: the type here `error` `disambiguation` `offline` can be overwritten by client website if they have the same class name with `!important`.

hueitan (commented):
It looks good to me, just one css concern.

src/preview.js
medied: Good call!

hueitan:
RTL style in disambiguation

image

src/preview.js
stephanebisson: I'm not sure what the "-template" part means. Why do we have `div.body-template` inside `div.body`?

medied: I think "-template" made more sense when we were including that CSS through `extend`.

What about replacing "-template" with "-state"?... or something along those lines

stephanebisson: Do we need 2 container levels? Can we have `<div class="wp-body wp-body-${type}">`?

medied: Actually you're right, check out https://github.com/wikimedia/wikipedia-preview/pull/81/commits/6d67140d5d2b7555694a7bd6812fce305b2b46f7. It abstracts it to only one container and now I think we can make a more appropriate use of `extend` with that `.wikipediapreview-body-non-regular`, similiar to what we're doing with `.license` in gallery.less (which by the way we should take care of [that TODO](https://github.com/wikimedia/wikipedia-preview/blob/master/style/gallery.less#L3) in either [T262371](https://phabricator.wikimedia.org/T262371) or [T262373](https://phabricator.wikimedia.org/T262373))

medied:
RTL style fixed now, good catch Huei

_mirror_
image

src/preview.js
```diff
@@ -26,88 +47,52 @@ const getPreviewHeader = ( lang, imageUrl = '' ) =>{
...
-							<div class="wikipediapreview-loading-body-line medium"></div>
-							<div class="wikipediapreview-loading-body-line smaller"></div>
-						</div>
-					</div>
-					<div class="wikipediapreview-loading-footer"></div>
-			</div>
-  `.trim()
+		const bodyContent = `
+      <div class="wikipediapreview-loading">
+        <div class="wikipediapreview-body">
```
hueitan, line 58: the other div container built in `getPreviewBody`

```
wikipediapreview-body wikipediapreview-body-${type}
```

Why in loading we need to resuse it here?

medied: Hmm you are correctly pointing out an inconsistency with the loading state HTML structure compared with the other states, if I'm understanding you well. How about https://github.com/wikimedia/wikipedia-preview/pull/81/commits/a4407940f9ad727c71dc9769234a75350967c47f? I'm updating the HTML as well as preview.less so it is more consistent with the rest of the states.

hueitan (approved):
LGTM now :+1:

PR #82, opened 2020-10-26
Fix IE11 styling

hueitan:
Phabricator Ticket : https://phabricator.wikimedia.org/T260781

Problem
===

IE11 has a lot of styling issues

Solution
===

 - [x] css value none doesn't existed in IE11
 - [x] Image Slider position

### DIFF
--- a/style/gallery.less
+++ b/style/gallery.less
@@ -11,7 +11,6 @@
 .wikipediapreview {
 	&-gallery {
 		display: flex;
-		display: -ms-flex;
 		height: 80px;
 		margin-top: 5px;
 		margin-right: 20px;
@@ -19,7 +18,6 @@
 
 		&-row {
 			display: flex;
-			display: -ms-flex;
 			overflow: scroll;
 		}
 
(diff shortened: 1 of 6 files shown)


hueitan:
this update is mainly for IE11, but still affect any other browser, it's worth to check again in chrome/firefox...

style/gallery.less
```diff
@@ -60,27 +58,24 @@
...
 		display: flex;
-		display: -ms-flex;
 		max-width: 100%;
 		height: 100%;
 	}
 
 	&-slider {
-		display: inline-flex;
-		display: -ms-inline-flex;
+		display: flex;
```
medied, line 66: Hopefully these won't conflict too much with #85 or #86. Working well in Firefox and Chrome for me right now!

medied:
So basically we should stay away from `unset` moving forward

medied (approved):

style/gallery.less
hueitan: good news 🎉 I have tested merging the others (only 1 conflict which easy to fix), works fine 💯

stephanebisson:
> So basically we should stay away from `unset` moving forward

See #87

stephanebisson (approved):

PR #85, opened 2020-10-30
T262371: Fix multiline image caption

medied:
https://phabricator.wikimedia.org/T262371

Opting for leaving the caption div as `position: absolute` as it was already. As suspected the ellipsis seem to be more troublesome than what it's worth, want to explore alternatives with Sudhanshu. More thoughts inline

### DIFF
--- a/src/gallery/slider.js
+++ b/src/gallery/slider.js
@@ -169,6 +169,20 @@ const clientWidth = window.innerWidth,
 		}
 	},
 
+	handleCaptionExpansion = ( item, forceClose = false ) => {
+		const captionText = item.querySelector( `.${prefixClassname}-item-caption-text` ),
+			expandCue = item.querySelector( `.${prefixClassname}-item-caption-expand-cue` ),
+			expanded = item.querySelector( '.expanded' )
+
+		if ( expandCue && expanded || forceClose && expandCue ) {
+			expandCue.classList.remove( 'expanded' )
+			captionText.style.maxHeight = '80px'
+		} else if ( expandCue ) {
+			expandCue.classList.add( 'expanded' )
+			captionText.style.maxHeight = '241px'
+		}
+	},
+
 	showImageAndInfo = ( index, refreshImage = false ) => {
 		const slider = parentContainer.querySelector( `.${prefixClassname}` ),
 			items = slider.querySelectorAll( `.${prefixClassname}-item` ),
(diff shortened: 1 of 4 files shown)


src/gallery/slider.js
```diff
@@ -79,12 +79,14 @@ const clientWidth = window.innerWidth,
 			},
 
 			author = mediaInfo.author ? mediaInfo.author : msg( lang, 'gallery-unknown-author' ),
-			link = mediaInfo.filePage
+			link = mediaInfo.filePage,
+			description = getImageDescription()
 
 		// @todo consider a wrapper container for all the image info?
```
medied, line 85: Did consider a wrapper but still not convinced, we could revisit in [T262373](https://phabricator.wikimedia.org/T262373)

src/gallery/slider.js
```diff
@@ -79,12 +79,14 @@ const clientWidth = window.innerWidth,
...
 		return `
 			<div class="${prefixClassname}-item-caption">
-				<div class="${prefixClassname}-item-caption-text">${getImageDescription()}</div>
+				${description.length > 95 ? `<div class="${prefixClassname}-item-caption-expand-cue"></div>` : ''}
```
medied, line 88: An idea to roughly estimate the number of lines the caption has.

stephanebisson: Did you test it with languages using other writing systems (Chinese, Arabic, Thai, Hebrew, etc)?

medied: I tested with Arabic and Thai from our demo site and seems to work fine. Indeed different languages behave slightly different and that's why using string length gives us only a rough estimate, but seems this could work. Still need to hear Sudhanshu's thoughts and reaction though. The `95` value is something we can tweak and fine-tune.

hueitan:
Missing this challenging part from the description

> If a caption is less than 4 lines long then show the entire caption. If a caption is longer than 4 lines then present it in 3 lines ending with ellipses and show the rounded rectangle which can be tapped to read the remaining part of it.

src/gallery/slider.js
medied: Synced with Sudhanshu today, I'm introducing https://github.com/wikimedia/wikipedia-preview/pull/85/commits/0a7e5f6ad56801bb5fb169f1879990a16611d4a0 to refine this a bit. Idea is to use both string length and screen width size to gauge better if the expand cue should be rendered or not. Will continue syncing/refining with him tomorrow.

hueitan: > Idea now is to use both string length and screen width size to gauge

it's actually look good to me with the magic number, if this pass the design, then I'm fine with it.

hueitan (approved):
Looks good to me

hueitan:
the caption div should be empty (with no text div inside), so the background image doesn't affect by this div background color

image

stephanebisson (commented):
This works well but it would be really useful to have some examples of small/medium/large captions in the Storybook.

medied:
> the caption div should be empty (with no text div inside), so the background image doesn't affect by this div background color

Good catch! 👏

> ...would be really useful to have some examples of small/medium/large captions in the Storybook.

Good idea, I have added some examples now

hueitan:
The storybook example looks good to me. 

Although it's better to use a different story for a different caption and allow changing the caption through the control, our `slider.js` component doesn't design for that due to the api request, it requires more code changing to the component itself.

So I'm fine with the current storybook approach for now.

PR #87, opened 2020-11-06
Make stylelint check for unsupported browser features

stephanebisson:
Stylelint will output warnings for CSS features that may not fully work in our target browsers.

### DIFF
--- a/browserslist
+++ b/browserslist
@@ -1,0 +1,3 @@
+last 2 Chrome versions
+last 2 Firefox versions
+IE 11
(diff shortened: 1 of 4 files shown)


browserslist
```diff
@@ -0,0 +1,3 @@
+last 2 Chrome versions
+last 2 Firefox versions
+IE 11
```
medied, line 3: Hmm this [`browserslist`](https://github.com/browserslist/browserslist) is so simple it looks curious

medied (approved):

stephanebisson:
I've created #90 to turn these warnings off.

PR #89, opened 2020-11-16
T266675 fix gallery storybook and add error state

hueitan:
Phabricator Ticket : https://phabricator.wikimedia.org/T266675#6623145

(1)
image

(2)
image

### DIFF
--- a/src/stories/gallery.stories.js
+++ b/src/stories/gallery.stories.js
@@ -69,6 +69,12 @@ export const Error = ( { lang, dir } ) => {
 		},
 		{
 			title: 'error.jpg'
+		},
+		{
+			title: 'error.jpg'
+		},
+		{
+			title: 'error.jpg'
 		}
 	]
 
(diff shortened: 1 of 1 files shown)


src/stories/gallery.stories.js
```diff
@@ -33,23 +33,42 @@ export const Fullscreen = ( { lang, dir } ) => {
...
+			title: 'File:English_Wikipedia_main_page_20011217.jpg'
+		}
+	]
+
+	showFullscreenGallery( mediaItems, '', lang, dir, container )
+
+	return container
+}
 
+export const Error = ( { lang, dir } ) => {
```
stephanebisson, line 64: This renders as just black for me.

medied: Same for me

hueitan: I remember sometimes I also see a blank error page, but I can't reproduce it now.

can you load the error state directly in http://localhost:6006/?path=/story/gallery--error? checkout the fullscreen story first then error.

hueitan: well, please try it again with the latest update

stephanebisson: It works for me now but it's weird. It should work with any number of images.

hueitan: yes, this is fixed here! https://github.com/wikimedia/wikipedia-preview/pull/93

PR #90, opened 2020-11-16
Allow features partially supported

stephanebisson:
Some web features that we rely on are only partially supported by IE11. The warnings emitted by the plugin are coming from a good place but are not relevant in this case.


hueitan (approved):
Thanks! It looks great!

PR #91, opened 2020-11-18
T268057 hide info when it is not available and handle refresh event

hueitan:
Phabricator : https://phabricator.wikimedia.org/T268057

Problem
===

The info icon is shown when the source link is not found from the api response

Solution
===

Simply hide the info element when there is no link given. However, it isn't idea when the page is **given offline then online** due to the reason how we design to refresh the image + info, therefore addition action during refresh image event.


medied (approved):
Looks good!

PR #94, opened 2021-02-24
Hooks

stephanebisson:
https://phabricator.wikimedia.org/T269934

Provide custom events hooks to allow 3rd party to build their own instrumentation when a popup is open and then the user navigates to read an article on wiki.

### DIFF
--- a/demo/articles/english.html
+++ b/demo/articles/english.html
@@ -55,7 +55,10 @@
         <script src="../wikipedia-preview.development.js"></script>
         <script>
             wikipediaPreview.init({
-                lang: 'en'
+                lang: 'en',
+                events: {
+                    onShow: console.log
+                }
             });
         </script>
         <!-- End of Scripts -->
--- a/demo/articles/english.html
+++ b/demo/articles/english.html
@@ -57,8 +57,12 @@
             wikipediaPreview.init({
                 lang: 'en',
                 events: {
-                    onShow: console.log,
-                    onReadOnWiki: console.log
+                    onShow: function(title, lang, type) {
+                        console.log(`Opening ${type} popup for article ${title} in ${lang}`)
+                    },
+                    onWikiRead: function(title, lang) {
+                        console.log(`Going to read article ${title} in ${lang} on Wikipedia`)
+                    }
                 }
             });
         </script>
--- a/demo/articles/english.html
+++ b/demo/articles/english.html
@@ -57,7 +57,8 @@
             wikipediaPreview.init({
                 lang: 'en',
                 events: {
-                    onShow: console.log
+                    onShow: console.log,
+                    onReadOnWiki: console.log
                 }
             });
         </script>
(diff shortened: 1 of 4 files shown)


src/index.js
```diff
@@ -5,19 +5,32 @@ import { createTouchPopup } from './touchPopup'
 import { renderPreview, renderLoading, renderError, renderDisambiguation, renderOffline } from './preview'
 import { getWikipediaAttrFromUrl, isTouch, getDir, isOnline } from './utils'
 
+const invokeCallback = ( events, name, ...params ) => {
```
hueitan, line 8:
```suggestion
const invokeCallback = ( events, name, params ) => {
```

this should be just `params` as it is already an array

demo/articles/english.html
```diff
@@ -55,7 +55,11 @@
         <script src="../wikipedia-preview.development.js"></script>
         <script>
             wikipediaPreview.init({
-                lang: 'en'
+                lang: 'en',
+                events: {
+                    onShow: console.log,
+                    onReadOnWiki: console.log
```
hueitan, line 61: onWikiRead?

hueitan (approved):
Looks good now

src/index.js
```diff
@@ -5,19 +5,32 @@ import { createTouchPopup } from './touchPopup'
 import { renderPreview, renderLoading, renderError, renderDisambiguation, renderOffline } from './preview'
 import { getWikipediaAttrFromUrl, isTouch, getDir, isOnline } from './utils'
 
+const invokeCallback = ( events, name, params ) => {
+	const callback = events && events[ name ]
+	if ( callback instanceof Function ) {
+		try {
+			callback.apply( null, params )
+		} catch ( e ) {
+			// we don't care about buggy 3rd party functions
```
medied, line 14: Would the errors be accessible to them? What if we at least `console.log` them in case they care?

src/index.js
```diff
@@ -11,7 +11,7 @@ const invokeCallback = ( events, name, params ) => {
 		try {
 			callback.apply( null, params )
 		} catch ( e ) {
-			// we don't care about buggy 3rd party functions
+			console.log('Error invoking Wikipedia Preview custom callback' e )
```
jpita, line 14: the build is failing because of the `e`  here, should this be fixed?

jpita:
issue found:
Steps
1-I move the mouse over a word 
2-move the mouse to the popup
3-move to the mouse to the word again

* Every time I repeat the steps 2 and 3 a new message appears. We should send only one message per "popup open", not when we over the mouse on the word

hueitan:
> issue found:
> Steps
> 1-I move the mouse over a word
> 2-move the mouse to the popup
> 3-move to the mouse to the word again
> 
> * Every time I repeat the steps 2 and 3 a new message appears. We should send only one message per "popup open", not when we over the mouse on the word

so, this "bug" ticket exist here https://phabricator.wikimedia.org/T259490

jpita:
> so, this "bug" ticket exist here https://phabricator.wikimedia.org/T259490
can be fixed here?

PR #95, opened 2021-02-24
External svg

stephanebisson:
https://phabricator.wikimedia.org/T273674

Remove the inline SVGs and load them from wikipedia.org/static


stephanebisson:
This is in draft until the images have been deployed to wikipedia static and this can be tested.

style/gallery.less
```diff
@@ -177,7 +177,7 @@
 					height: 44px;
 					width: 44px;
 					display: flex;
-					background-image: inline-svg( '../images/attribution-icon-info.svg' );
+					background-image: url( https://wikipedia.org/static/images/wikipedia-preview/attribution-icon-info.svg );
```
Krinkle, line 180: The canonical domain name is `www.wikipedia.org`, avoid redirects or staleness.

stephanebisson: Thanks, will update it

style/gallery.less
```diff
@@ -473,13 +473,13 @@
 
 				&-button {
 					&.previous {
-						background-image: inline-svg( '../images/arrow-right.svg' );
+						background-image: url( https://wikipedia.org/static/images/wikipedia-preview/arrow-right.svg );
```
Krinkle: These are cached for 1 year unconditionally. That might be fine, but it might also be good to include a `?something` here to indicate that these are treated as permalinks, and thus to give you a self-documented and discoverable way to invalidate it in case it ever needs to change.

In MW we auto-append those based on `md5(bytes).slice(0, 5)`, I don't know if the build step here can easily support that, but otherwise a simple incrementing number local to the file, or today's date might work.

stephanebisson: Any of those options would be easy to set up here.

I played around with a plugin here: 68b286

PR #96, opened 2021-03-01
Ignore hover for link with open popup

stephanebisson:
https://phabricator.wikimedia.org/T259490

When a target element already has an open popup attached to it, ignore subsequent hovers.

This is causing multiple hook onOpen events for the same popup in #94.


medied (approved):
Neat use of `refresh`

hueitan (approved):
Nice!

PR #98, opened 2021-03-16
T276281: Improve height setting when expanding down

medied:
https://phabricator.wikimedia.org/T276281

## Problem 

When preview popup expands down with limited screen space, it's expanding too far down the screen and overflows the available screen space. 

## Solution

Let `setPreviewMaxHeight` also decide what's the most appropriate height when expanding down, similar to how we do it when expanding up.


hueitan (approved):
Looks good to me.

PR #99, opened 2021-03-21
Cypress wikimedia

Aditya-1295:
Hi , @jpita . I want to take part in Gsoc , i am intrested  for the idea - GSOC - Create cypress tests for wikipedia-preview .


Aditya-1295:
And how can i solve the update demos test ?

jpita:
if you want to participate, add a comment on this task https://phabricator.wikimedia.org/T277284

Aditya-1295:
> you can't change anything in this file, remove all the changes in this file.
> instead of data-link you need to find an already existing locator https://docs.cypress.io/api/commands/get.html#Selector



> * you need to format this file with proper indentation
> * you are not testing anything here https://docs.cypress.io/guides/references/assertions.html
> * avoid using fix time waits https://docs.cypress.io/guides/references/best-practices.html#Unnecessary-Waiting
> * instead of using a for loop for each article, create one test for each article and an assertion for each as well

Thankyou for the review , I will try to write tests accordingly.

PR #104, opened 2021-04-07
T278456: Add wix integration instructions

medied:
https://phabricator.wikimedia.org/T278456

### Objective

Add detailed instructions on how to use Wikipedia Preview in Wix Sites.

### DIFF
--- a/wix.md
+++ b/wix.md
@@ -35,4 +35,4 @@ Then to initialize:
   6. By default this code should be marked as "Essential", you can ensure this is the case under the Code Type tab.
   7. Click "Apply"
 
-  You are now ready to start using Wikipedia Preview in your Wix Site. Simply [add hyperlinks](https://support.wix.com/en/article/wix-editor-adding-a-text-hyperlink) through the Wix Editor with the Wikipedia article link you would like to use on a specific word. Once you publish your site to the updated version, Wikipedia Preview popups should appear on link hover (desktop) or touch (mobile).
+You are now ready to start using Wikipedia Preview in your Wix Site. Simply [add hyperlinks](https://support.wix.com/en/article/wix-editor-adding-a-text-hyperlink) through the Wix Editor with the Wikipedia article link you would like to use on a specific word. Once you publish your site to the updated version, Wikipedia Preview popups should appear on link hover (desktop) or touch (mobile).
--- a/wix.md
+++ b/wix.md
@@ -1,0 +1,38 @@
+# Wix Sites integration
+
+You can use Wikipedia Preview in your [Wix Site](https://www.wix.com/) if you have any premium account by [embedding custom code to your site](https://support.wix.com/en/article/embedding-custom-code-to-your-site). Follow these simple instructions to get started and bring the sum of all knowledge to your readers.
+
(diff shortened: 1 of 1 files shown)


wix.md
```diff
@@ -0,0 +1,38 @@
+# Wix Sites integration
+
+You can use Wikipedia Preview in your [Wix Site](https://www.wix.com/) if you have any premium account by [embedding custom code to your site](https://support.wix.com/en/article/embedding-custom-code-to-your-site). Follow these simple instructions to get started and bring the sum of all knowledge to your readers.
+
+First head to your site's [dashboard settings](https://www.wix.com/my-account/site-selector/?buttonText=Open%20Settings&title=Select%20a%20Site&autoSelectOnSingleSite=true&actionUrl=https://www.wix.com/dashboard/{{metaSiteId}}/settings) to load and initialize Wikipedia Preview:
+
+To load:
+  1. Click the Custom Code tab under Advanced Settings
+  2. Click either "Add Custom Code" at the top right or the "Add Code" button in the "Head" box
```
stephanebisson, line 9: This file should be loaded at the end of the body so it doesn't slow down initial rendering. It looks like both load and init can be included in the same code block for simplicity.

medied: I like that simplicity. Although given that we need to "Load code on each new page" for `init` to trigger, if we include both in the same code block I'm seeing the `unpkg` get added many times in the HTML. Is that acceptable?

image

stephanebisson: Not ideal. Let's create 2 blocks but both at the end of the body.

medied:
I'm afraid this is not quite sorted out yet: after further testing I'm seeing there are issues with the preview popup appearing on blog post pages. 

The loading script is working fine, it’s the `init` script that’s misbehaving: it works fine on home page but in post pages it only works if you do a hard refresh in the post url directly (for example: https://www.wixwikipediapreviewtest.com/post/the-guide-to-successful-photography-blogging). I have tried different combinations of the custom code `init` placement (head, body start, body end, "load once", "load on each new page”), but no luck yet.

stephanebisson:
> The loading script is working fine, it’s the `init` script that’s misbehaving: it works fine on home page but in post pages it only works if you do a hard refresh in the post url directly...

Maybe [$w.onReady](https://www.wix.com/velo/reference/$w/onready) can help

medied:
> Maybe $w.onReady can help

Issue with `$w.onReady` is that I'm not able to access our `wikipediaPreview` compoment from within `$w.onReady`. This seems to be on purpose (I suppose as a security measure?). These are some things I've tried so far:

- Grab the [script by id](https://www.wix.com/velo/reference/$w/$w)
- Communicate from script to page code (`$w.onReady`) [using messages](https://support.wix.com/en/article/velo-working-with-the-html-iframe-element#messaging-1)
- Using the localStorage as leverage, as in storing inside script and then grabbing within `$w.onReady`

But the `wikipediaPreview` component has not been accessible from `$w.onReady` in any case, it's always either an empty object or array or `nulll`.

medied:
Dead-end, closing

https://www.wix.com/velo/forum/coding-with-velo/bringing-wikipedia-to-wix-sites
https://phabricator.wikimedia.org/T278456#7159957

PR #105, opened 2021-04-09
T278377: Add version logging method

medied:
https://phabricator.wikimedia.org/T278377

### Objective

Wikipedia Preview should print its version number to the javascript console, as that is useful information when a partner requests support.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -6,16 +6,21 @@ import { renderPreview, renderLoading, renderError, renderDisambiguation, render
 import { getWikipediaAttrFromUrl, isTouch, getDir, isOnline } from './utils'
 
 const invokeCallback = ( events, name, params ) => {
-	const callback = events && events[ name ]
-	if ( callback instanceof Function ) {
-		try {
-			callback.apply( null, params )
-		} catch ( e ) {
-			// eslint-disable-next-line no-console
-			console.log( 'Error invoking Wikipedia Preview custom callback', e )
+		const callback = events && events[ name ]
+		if ( callback instanceof Function ) {
+			try {
+				callback.apply( null, params )
+			} catch ( e ) {
+				// eslint-disable-next-line no-console
+				console.log( 'Error invoking Wikipedia Preview custom callback', e )
+			}
 		}
+	},
+
+	version = () => {
+	/* eslint-disable-next-line no-undef, no-console */
+		console.log( `Wikipedia Preview - ${APP_VERSION} (${CIRCLE_SHA1})` )
 	}
-}
 
 let currentPopupId
 
@@ -146,6 +151,8 @@ function init( {
 			}
 		)
 	}
+
+	version()
 }
 
 export { init }
--- a/src/index.js
+++ b/src/index.js
@@ -150,4 +150,4 @@ function init( {
 
 version()
 
-export { init }
+export { init, version }
--- a/src/index.js
+++ b/src/index.js
@@ -3,24 +3,19 @@ import { customEvents } from './event'
 import { createPopup } from './popup'
 import { createTouchPopup } from './touchPopup'
 import { renderPreview, renderLoading, renderError, renderDisambiguation, renderOffline } from './preview'
-import { getWikipediaAttrFromUrl, isTouch, getDir, isOnline } from './utils'
+import { getWikipediaAttrFromUrl, isTouch, getDir, isOnline, version } from './utils'
 
 const invokeCallback = ( events, name, params ) => {
-		const callback = events && events[ name ]
-		if ( callback instanceof Function ) {
-			try {
-				callback.apply( null, params )
-			} catch ( e ) {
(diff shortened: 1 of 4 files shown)


webpack.config.js
```diff
@@ -38,7 +40,11 @@ const config = {
       new BundleAnalyzerPlugin({
         analyzerMode: 'disabled',
         generateStatsFile: true
-      })
+      }),
+      new webpack.DefinePlugin({
+        APP_VERSION: JSON.stringify(JSON.parse(fs.readFileSync('./package.json')).version),
+        CIRCLE_SHA1: JSON.stringify(process.env.CIRCLE_SHA1 || '-')
```
medied, line 46: Let's please review this: is that all there is to the git hash? How can we test this?

https://circleci.com/docs/2.0/env-vars/

src/index.js
```diff
@@ -6,16 +6,21 @@ import { renderPreview, renderLoading, renderError, renderDisambiguation, render
...
+			} catch ( e ) {
+				// eslint-disable-next-line no-console
+				console.log( 'Error invoking Wikipedia Preview custom callback', e )
+			}
 		}
+	},
+
+	version = () => {
+	/* eslint-disable-next-line no-undef, no-console */
+		console.log( `Wikipedia Preview - ${APP_VERSION} (${CIRCLE_SHA1})` )
```
medied, line 22: Would it be useful to add the version as 'v1.1.1' or 'version 1.1.1'?

src/index.js
```diff
@@ -6,16 +6,21 @@ import { renderPreview, renderLoading, renderError, renderDisambiguation, render
...
 
 const invokeCallback = ( events, name, params ) => {
-	const callback = events && events[ name ]
-	if ( callback instanceof Function ) {
-		try {
-			callback.apply( null, params )
-		} catch ( e ) {
-			// eslint-disable-next-line no-console
-			console.log( 'Error invoking Wikipedia Preview custom callback', e )
+		const callback = events && events[ name ]
```
stephanebisson, line 9: Not sure why the indentation changed here.

src/index.js
stephanebisson: I prefer the longer form (version 1.1.1)

webpack.config.js
stephanebisson, line 46: CIRCLE_SHA1 is unlikely to work here. Unlike the kaios app, this package is not produced in CI but locally. 

You can use `git rev-parse --short HEAD` to get the sha1.

stephanebisson (changes requested):

src/index.js
hueitan: questions 

is it better to have this in the util?
when should the version number printed? init() or script loaded?

stephanebisson: I think it should print on script load just to know which version is included. The debug mode will provide info about init invocations.

medied: Alright, will invoke `version` outside of `init` within index.js. And I'll just make it a util as well, why not

src/index.js
medied: It's the linter autocorrecting because of https://eslint.org/docs/rules/one-var. Actually moving `version` to util should restore the original indentation

stephanebisson: Btw, `one-var` is being dropped in the official MW style. We'll need to update our dependency soon.

medied: Oh nice! I'm actually not a huge fan of that rule so I'm not mad about it being dropped. Would you like me to create a bare task for that as a reminder?

webpack.config.js
medied: Awesome, thanks! Updated now

hueitan (approved):
Looks good!

stephanebisson (changes requested):
From phab:
> Add a version() function to the main object that prints the message above

This means that I should be able to call `wikipediaPreview.version()` in case the original message is lost in spam.

> Update the documentation in the README accordingly

Document the fact that the method above can be called on the main object.

medied:
> This means that I should be able to call wikipediaPreview.version() in case the original message is lost in spam.

Oh gotcha, agree it would be useful to be able to call `wikipediaPreview.version()`. This means `version` needs to be included as an `export` in index.js (which is the entry file specified in webpack.config.js). If we leave `version` in utils.js as is right now, import it in index.js and then export it from index.js, `wikipediaPreview.version()` works as expected... but does it seem weird to define it in one place and export twice like that? 

Will commit that so you all can see it in action and we can take it from there, wondering if it still makes sense to include it as a util though

stephanebisson:
I don't think it's weird to take it from util and export it again from index. It declutters index a little and that's important right now.

stephanebisson (approved):

src/index.js
stephanebisson: > Would you like me to create a bare task for that as a reminder?

por favor

PR #108, opened 2021-04-13
T278381 Debug Mode

hueitan:
Phabricator Ticket : https://phabricator.wikimedia.org/T278381

Debug Mode

 - [x] Update README
 - [x] Show the relevant init message

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -3,7 +3,7 @@ import { customEvents } from './event'
 import { createPopup } from './popup'
 import { createTouchPopup } from './touchPopup'
 import { renderPreview, renderLoading, renderError, renderDisambiguation, renderOffline } from './preview'
-import { getWikipediaAttrFromUrl, isTouch, getDir, isOnline, version } from './utils'
+import { getWikipediaAttrFromUrl, buildWikipediaUrl, isTouch, getDir, isOnline, version } from './utils'
 
 const invokeCallback = ( events, name, params ) => {
 	const callback = events && events[ name ]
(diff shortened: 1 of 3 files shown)


hueitan:
Example screenshot from the console

image

src/index.js
```diff
@@ -142,10 +151,37 @@ function init( {
...
+						lang: matches.lang
+					} )
 				}
 			}
 		)
 	}
+
+	if ( debug ) {
+		/* eslint-disable no-console */
+		console.group( 'Wikipedia Preview [debugger]' )
```
stephanebisson, line 167: `debugger` exists as a specific instruction in javascript, maybe use `[debug mode]` instead to avoid confusion.

src/index.js
```diff
@@ -142,10 +151,37 @@ function init( {
...
+		console.log( `searching for the content in language "${globalLang}" within the selector of "${selector}"` )
```
stephanebisson, line 168: Suggestion: Searching for "${selector}" inside "${root}"... Found: ${foundSelectorLinks.length}

src/index.js
```diff
@@ -142,10 +151,37 @@ function init( {
...
+		console.group( `Total links found in the given selector of "${selector}" : ${foundSelectorLinks.length}` )
+		foundSelectorLinks.forEach( ( link, index ) => {
+			console.log( index + 1, `text "${link.text}", title "${link.title}", lang : "${link.lang}"` )
+		} )
+		console.groupEnd()
+		if ( detectLinks ) {
+			console.group( `Total links found in the hyperlink : ${foundDetectLinks.length}` )
```
stephanebisson, line 175: Suggestion: Searching for links to Wikipedia... Found: ${foundDetectLinks.length}

src/index.js
```diff
@@ -142,10 +151,37 @@ function init( {
...
 		)
 	}
+
+	if ( debug ) {
+		/* eslint-disable no-console */
+		console.group( 'Wikipedia Preview [debugger]' )
+		console.log( `searching for the content in language "${globalLang}" within the selector of "${selector}"` )
+		console.group( `Total links found in the given selector of "${selector}" : ${foundSelectorLinks.length}` )
+		foundSelectorLinks.forEach( ( link, index ) => {
+			console.log( index + 1, `text "${link.text}", title "${link.title}", lang : "${link.lang}"` )
```
stephanebisson, line 171: What do you think of including the article URL here (we have a util for that)? It could be something like:
`<text of the target> -> <url of the article>`

Example: `Cat -> https://es.wikipedia.org/wiki/Gato`

stephanebisson (changes requested):
Looks good! Just need to work on the wording a little.

src/index.js
hueitan: yeah, this sounds good.

stephanebisson (changes requested):
Could you add a param to the util so we don't include the ?wpprov argument in the conosole output?

stephanebisson (approved):
LGTM

medied (approved):
Yea this is pretty cool, well done Huei. I think this is a feature that would be interesting to revisit once a partner/developer actually uses it and feedback is provided. I can imagine maybe it being useful to have a `wikipediaPreview.debug()` similar to `wikipediaPreview.version()`, for example. 

Looks good!

PR #109, opened 2021-04-21
Smoothen gallery closing animation

medied:
https://phabricator.wikimedia.org/T269854

### Problem

The gallery caption transition from expanded to collapsed is awkward because there's a `transition` rule for the caption text but not for the caption expand cue. 

### Solution

I'm updating the collapsed state of the expand cue with an SVG (`svgo` ready) so that we can set a `transition` rule for the expand cue itself to go along with the caption text. This makes the overall animation smoother.

### DIFF
--- a/images/caption-cue.svg
+++ b/images/caption-cue.svg
@@ -1,0 +1,1 @@
+<svg width="37" height="3" fill="none" xmlns="http://www.w3.org/2000/svg"><path d="M0 1h36.5M0 2h36.5" stroke="#fff"/></svg>
--- a/style/gallery.less
+++ b/style/gallery.less
@@ -85,24 +85,25 @@
 
 				&-expand-cue {
 					width: 37px;
-					height: 2px;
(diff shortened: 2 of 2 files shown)


hueitan:
I can't spot the difference with my 👀 , can you do some gif or pics before/after comparison?

stephanebisson (commented):
It's hard to see. the animation might be smoother but the task is about changing the order on collapse: it should 1) lower the caption box and 2) turn the down arrow into a straight line.

medied:
Yea it's hard to see, it's best to compare with the before and after right next to each other. Sharing a view now via Slack to bring design eyes into the conversation. I've tried a few things but the order on collapse seems to be behind a 'browser painting' wall, I'd like to revisit that once we get design input

hueitan:
why this is merged?

jpita:
> why this is merged?

I only saw the error after the merge

PR #110, opened 2021-04-29
T281299 Image Case

hueitan:
Phabricator Ticket : https://phabricator.wikimedia.org/T281299

The wikipedia preview doesn't work in image case

```html
<a href="https://en.wikipedia.org/wiki/Cat/">
   <img src="./cute-cat.jpg"/>
</a>
```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -42,7 +42,7 @@ function init( {
 
 		const popupId = Date.now()
 		const { currentTarget } = refresh ? last : e
-		const title = refresh ? last.title : currentTarget.getAttribute( 'data-wp-title' ) || currentTarget.textContent
+		const title = refresh ? last.title : decodeURIComponent( currentTarget.getAttribute( 'data-wp-title' ) || currentTarget.textContent )
 		const localLang = refresh ? last.lang : currentTarget.getAttribute( 'data-wp-lang' ) || globalLang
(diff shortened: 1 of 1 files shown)


hueitan:
do not merge, i need more testing and verification of this attribute.

stephanebisson:
What's the documented difference between `target` and `currentTarget`?

hueitan:
According to the mdn, it seems like `currentTarget` is the right one we need

> The target property of the Event interface is a reference to the object onto which the event was dispatched. It is different from **Event.currentTarget** when **the event handler is called** during the bubbling or capturing phase of the event.

> The **currentTarget** read-only property of the Event interface identifies the current target for the event, as the event traverses the DOM. It always refers to the element to which the **event handler has been attached**, as opposed to Event.target, which identifies the element on which the event occurred and which may be its descendant.

hueitan:
This works well in IE11 / Edge / Chrome too. Remember to test the normal case and the image case in this change.

stephanebisson:
Would be nice to have an example on the demo page or some way in the storybook to test this.

stephanebisson:
This PR shows no changes now. Was it lost in the merge with main?

hueitan:
> This PR shows no changes now. Was it lost in the merge with main?

hmmm, i think the merge and conflict remove the previous changes!

PR #112, opened 2021-04-29
T280061: Update eslint-config-wikimedia to 0.20

medied:
https://phabricator.wikimedia.org/T280061

### Objective

Updates linter to 0.20. Please review and test carefully, more specific comments in-line. You should `rm -rf node_modules` and `npm install` first

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -17,6 +17,7 @@ module.exports = {
     sourceType: 'module'
   },
   rules: {
-    semi: [2, "never"]
+    semi: [2, "never"],
+    'no-nonoctal-decimal-escape': 'off'
   }
 }
--- a/src/index.js
+++ b/src/index.js
@@ -43,10 +43,9 @@ function init( {
 		const popupId = Date.now()
 		const { target } = refresh ? last : e
 		const title = refresh ? last.title : target.getAttribute( 'data-wp-title' ) || target.textContent
-		// eslint-disable-next-line no-shadow
-		const lang = refresh ? last.lang : target.getAttribute( 'data-wp-lang' ) || globalLang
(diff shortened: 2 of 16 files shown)


.eslintrc.js
```diff
@@ -17,6 +17,7 @@ module.exports = {
     sourceType: 'module'
   },
   rules: {
-    semi: [2, "never"]
+    semi: [2, "never"],
+    'no-nonoctal-decimal-escape': 'off'
```
medied, line 21: No idea how to address this error other than here, any thoughts?

src/index.js
```diff
@@ -28,94 +28,96 @@ function init( {
...
+	const foundSelectorLinks = []
+	const foundDetectLinks = []
+
+	const showPopup = ( e, refresh = false ) => {
+		e.preventDefault()
+
+		const popupId = Date.now()
+		const { target } = refresh ? last : e
+		const title = refresh ? last.title : target.getAttribute( 'data-wp-title' ) || target.textContent
+		// eslint-disable-next-line no-shadow
```
medied, line 46: `lang` is technically declared already from method arguments. Let me know if we want to update the name of one or other, it'd likely have repercussions in other parts of the code

stephanebisson: We should fix that by renaming `lang` to something else so it doesn't shadow the `lang` argument from the outer function.

.eslintrc.js
stephanebisson: What is this rule?

src/index.js
medied: How about `localLang` to parallel the `globalLang` we already have in there? See https://github.com/wikimedia/wikipedia-preview/pull/112/commits/b4fde023b472ec1d9522ee1887cd8f6799e27ae6

PR #114, opened 2021-05-11
T263598 keyboard shortcuts on gallery

hueitan:
Phabricator Ticket : https://phabricator.wikimedia.org/T263598

 - [x] LTR / RTL cases
 - [x] different browser

### DIFF
--- a/src/gallery/event.js
+++ b/src/gallery/event.js
@@ -1,0 +1,16 @@
+let eventListenerStack = []
+
+const addEventListener = ( target, type, listener, options = undefined ) => {
+	target.addEventListener( type, listener, options )
+	eventListenerStack.push( [ target, type, listener, options ] )
+}
+
+const clearAllEventListener = () => {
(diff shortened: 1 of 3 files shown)


src/gallery/fullscreen.js
```diff
@@ -34,6 +36,15 @@ const showFullscreenGallery = (
 		hideFullscreenGallery( container )
 	} )
 
+	addEventListener( window, 'keydown', event => {
```
stephanebisson, line 39: You can destructure to `({ key })` instead of `event`

stephanebisson (approved):

src/gallery/event.js
```diff
@@ -0,0 +1,16 @@
+let eventListenerStack = []
```
medied, line 1: Should all gallery event listeners be included in this new array? The ones in `applyGestureEvent` for example? Mainly asking because that would overlap the image zoom feature I'm working on.

PR #116, opened 2021-05-17
T269787: Gallery zoom feature

medied:
https://phabricator.wikimedia.org/T269787

### Objective

Introduce zooming ability to the gallery. 

### Highlights

* The core of the pinching gesture comes directly from [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events/Pinch_zoom_gestures), it is based on pointer events
* The new _gestures.js_ includes all gallery gestures (both the existent sliding logic and the added zooming gesture), so you should recognize the logic in methods such as `slideStart`, `slideMove`, `slideEnd`. 
* This means that _slider.js_ gets a bit of a clean up/refactor: now `applyGestureEvents` acts as the orchestrating method that decides when and how should the gesture methods be applied so they all dance together smoothly. 
* Desktop view was left alone as it's out of scope, but you will notice the single and double tap work. 

---

_TODO_
- [x] Allow zoom-in and zoom-out with pinching gestures 
- [x] Add double-tap listener 
- [x] Allow scrolling/panning when image is zoomed in
- [x] More testing and clean up

---

_REFERENCES_

https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events/Pinch_zoom_gestures
https://developer.mozilla.org/en-US/docs/Web/CSS/transform
https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/scale()
https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/translate3d()
https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events#touch-action_css_property
https://dev.to/dailydevtips1/console-log-on-chrome-for-ios-i66


hueitan:
fyi, I'm testing the zoom/slide feature in google images, it uses `transform: translate3d() scale()` and super smooth!
image

medied:
> fyi, I'm testing the zoom/slide feature in google images, it uses `transform: translate3d() scale()` and super smooth!

Thank you for pointing this out Huei! As you can tell by [the commit](https://github.com/wikimedia/wikipedia-preview/pull/116/commits/54f295b374264a4017dd0b7b79c92381bdbe58ca) following your comment, `translate3d` proved to be useful here. I got up to two sub-branches trying to figure out scrolling.

```
Eduardos-MacBook-Pro:wikipedia-preview medied$ git branch -a
* T269787-gallery-zoom
  T269787-gallery-zoom-scrolling
  T269787-gallery-zoom-scrolling-2
```

stephanebisson (changes requested):
Code is very clean but I found a bug: zoom in and swipe the image up or down and it can go outside the viewport in a way that is not needed. It's OK for the top of the image to go up outside the viewport if it allows you to see the bottom of the image. When the bottom of the image is visible, it shouldn't go higher. There's no point bringing the bottom of the image to the middle of the screen.

src/gallery/gestures.js
```diff
@@ -0,0 +1,208 @@
...
+		zoomedIn = false
+	} else {
+		image.style.transform = `scale(${scaleMax})`
+		zoomedIn = true
+	}
+}
+
+const zoomStart = ( e ) => {
+	const image = grabImageFromEvent( e )
+	const imageStyle = window.getComputedStyle( image )
```
hueitan, line 79: When touching the cue, I got this error.

src/gallery/gestures.js
```diff
@@ -0,0 +1,208 @@
...
+const grabScaleFromTransform = ( transform ) => {
+	return Number( transform.slice( transform.indexOf( 'scale' ) + 6, -1 ) )
+}
+
+const isInvalidEvent = ( e, items, current, captionText, prefixClassname ) => {
+	return e.target.className === captionText &&
+			items[ current ].querySelector( `.${prefixClassname}-item-caption-expand-cue` ) ||
+			e.target.className === captionText ||
+			e.pointerType !== 'touch' ||
+			e.target.className.indexOf( 'slider-button' ) > -1
```
hueitan, line 34: `e.target.className === captionText` is duplicated in here.

correct me if i'm wrong, it seems like ``items[ current ].querySelector( `.${prefixClassname}-item-caption-expand-cue` )`` doesn't really needed here

maybe we can make this method better like given

```js
// example
const isInvalidevent = ( e, invalidClassNameArray ) => {
    return e.pointerType !== 'touch' || // first detect touch
        'algorithm' // e.target.className or classList contains one from the invalidClassNameArray
}
```

medied: Yea good idea! I sort of ported this method over from the previous logic and it was in much need of a refactor. Is it better now?

hueitan: Nice, exactly what I mean here.

src/gallery/gestures.js
```diff
@@ -126,8 +126,9 @@ const zoomScroll = ( e, renderNext, items, current ) => {
 	const image = grabImageFromEvent( e )
 	const transform = image.style.transform
 	const scale = transform ? grabScaleFromTransform( transform ) : scaleMin
+	const isImgLandscape = image.naturalHeight <= image.naturalWidth
 	const horizontalLimit = clientWidth / 2
-	const verticalLimit = clientHeight / 2
+	const verticalLimit = isImgLandscape ? clientHeight / 8 : clientHeight / 2
```
hueitan, line 137: can you explain this line? especially the `8` magic number 🤔

src/gallery/slider.js
```diff
@@ -328,14 +313,25 @@ const onShowFn = () => {
...
 
 	sliderContainer.addEventListener( 'click', ( e ) => {
 		if ( e.target.className === `${prefixClassname}-item` ||
 				e.target.tagName === 'IMG' ) {
-			toggleFocusMode()
+			if ( !tapped ) {
+				tapped = setTimeout( () => {
+					tapped = null
+					toggleFocusMode()
+				}, 300 )
```
hueitan, line 328: previously there wasn't timeout 300 for the single tap toggling, is there any way to not timeout it? but still keep the behaviour of double tap doesn't toggle it?

or, should we keep the caption visible when in zoom mode?

src/gallery/gestures.js
medied: I ended up leaving the image `transform-origin` as the default `center`. What that means is that once the image is enlarged, all the `translate3d` coordinates start from the center as zero (`(0, 0, 0)`). If we visualize that for the `horizontalLimit` for example, then from the center to right/left the distance "available" to scroll within the viewport is the `clientWidth` divided by two (one half to the right, one to left). 

Now let's take that same logic to the `verticalLimit`. If the image is a portrait, then `clientHeight / 2` will give us all of the available distance from the top/bottom halfs to scroll freely and see all of the zoomed in image. However if the image has a landscape orientation then using all of the distance available results in the unwanted behavior that Stephane pointed out: the bottom or top of the image could reach the middle of the screen leaving an awkward, empty space visible. The idea I'm proposing here there is to set the vertical limit to a lower value if the image is landscape so that we can restrain how much distance is available to scroll. Why 8? It's just the value that worked better after trying out different numbers (so we could find tune that number if needed). Does that make sense?

src/gallery/slider.js
medied: I tried originally with https://developer.mozilla.org/en-US/docs/Web/API/Element/dblclick_event but it doesn't support mobile unfortunately. After that [a search pointed me](https://stackoverflow.com/questions/24058241/touch-device-single-and-double-tap-events-handler-jquery-javascript) to this idea using timeouts... so would have to think about this a bit more if we wanted to not use timeouts. Are you concerned that using a timeout here might make thing more complicated down the road? 

> or, should we keep the caption visible (hidden?) when in zoom mode?

We can confirm with Sudhanshu. Last time I talked with him he did mention the behavior he is expecting is that a single tap can toggle the focus mode regardless of whether the image is zoomed in or not

hueitan: > single tap can toggle the focus mode regardless of whether the image is zoomed in or not

Thanks. Then I think there is nothing we can do if we use one click event to detect either single or double tap without timeout. I'm closing this thread conversation now unless others comment on this or have better idea.

hueitan (commented):
I don't see any other problems right now beside the console error when click the clue. I think it's alright to let QA and design to test and fine tune the behaviour.

hueitan (approved):
This is here for a while since last comment, let's move on and let qa design to review and decide if any changes needed.

PR #117, opened 2021-06-02
remove dev of bufferutil and canvas

hueitan:
I don't remember when we start having this, and where we use it.

any idea?


hueitan:
it seems this is for the storybook.

PR #118, opened 2021-06-02
Remove eslint rule from storybook

hueitan:
UPDATE: Storybook reads the eslint rule config from the root.

----

Silent error during storybook built

**Question**, why do we have eslint for storybook? can **it use the root eslint instead**?

image


stephanebisson:
I don't know if there was a specific reason for a distinct eslint config. Have you tried removing it?

medied:
Hmm would be neat if it's only one eslint config to take care of. For context, this is when we added the rule to omit the nonoctal decimal escape check: https://github.com/wikimedia/wikipedia-preview/pull/112#discussion_r623314252

hueitan:
Yes, removing works. We set the storybook uses the root webpack config, therefore it uses the same eslint conf.

PR #119, opened 2021-06-03
T282717 Add target stories for hyperlink, image and text

hueitan:
Phabricator Ticket: https://phabricator.wikimedia.org/T282717

 - [x] add target stories
 - [x] (Out ot scope, move to another ticket) add image in the demo page as the example

### DIFF
--- a/src/stories/target.stories.js
+++ b/src/stories/target.stories.js
@@ -1,0 +1,52 @@
+import { buildWikipediaUrl } from '../utils'
+import { init } from '../index'
+
+export default {
+	title: 'Target Cases',
+	argTypes: {
+		lang: {
+			name: 'Language',
+			defaultValue: 'en',
+			control: {
+				type: 'select',
+				options: [ 'en', 'fr', 'hi', 'ks', 'he' ]
+			}
+		},
+		title: {
+			name: 'Article Title',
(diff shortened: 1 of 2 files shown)


src/stories/target.stories.js
```diff
@@ -0,0 +1,51 @@
...
+			control: 'text'
+		}
+	}
+}
+
+export const Hyperlink = ( { lang, title } ) => {
+	const container = document.createElement( 'div' )
+	const template = `<a href="${buildWikipediaUrl( lang, title, true, false )}">${title} (${lang})</a>`
+	container.innerHTML = template
+	container.onmouseover = () => {
```
stephanebisson, line 27: Have you tried calling the init function directly instead of putting it in the mouseover handler? The `root` option should be the `container` variable.

src/stories/target.stories.js
```diff
@@ -0,0 +1,45 @@
+import { buildWikipediaUrl } from '../utils'
+import { init } from '../index'
+
+export default {
+	title: 'Target Cases',
```
stephanebisson, line 5: Maybe name that `Targets` without the "cases"

stephanebisson (approved):

PR #126, opened 2021-07-12
T285667: Add styling for unordered lists 

medied:
https://phabricator.wikimedia.org/T285667 

We were missing styling support for unordered lists (aka bullet points).  Added padding so the bullet points line up with the paragraph above, as well as the same paragraph line height.

Before | After
------  | ------
image | image

### DIFF
--- a/style/preview.less
+++ b/style/preview.less
@@ -94,6 +94,12 @@
 			padding: 10px 20px;
 		}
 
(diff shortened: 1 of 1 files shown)


hueitan:
The margin-right doesn't look right

hueitan (approved):

PR #128, opened 2021-07-16
Week-1 task Completed (#122)

jpita:
* Week-1 task Completed

* PR Review changes

* PR Review Changes

* PR Review changes

* Layout Testing Changes

* Offline Testing Functions

* Delete expand-spec.js

* add eslint package and run eslint

* small fix, timeout increase

* skiping failing tests

* Update config.yml

add cypress tests

* PR Review Changes added

* Removed Waits

* Changes check Preview function to objects

* Changes check Preview function to objects

* check.js Removed and helpers folder created for network.js

* Added page.js in Page-Object

* final changes

* bump cypress version, make code more clear

* package.lock

* package.json changed

* Package.json Changed

* package.json changed

* config.yml changed

* adding waits

* adding waits

Co-authored-by: Gabriel Pita <dvpita@gmail.com>


package.json
```diff
@@ -14,7 +14,8 @@
     "dev": "webpack-dev-server --mode development --host=0.0.0.0 --disable-host-check",
     "storybook": "start-storybook -p 6006",
     "build-storybook": "build-storybook",
-    "wba": "npm run build:prod && npx webpack-bundle-analyzer dist/stats.json"
+    "wba": "npm run build:prod && npx webpack-bundle-analyzer dist/stats.json",
+    "cypress:run": "cypress run"
```
hueitan, line 18: remember to include running eslint in cypress folder, you can either modify

`"test:fix-lint": "eslint test cypress --fix"`

or use a new script

`"cypress:test:fix-int": "eslint cypress --fix"`

.eslintrc.js
```diff
@@ -18,6 +19,12 @@ module.exports = {
   },
   rules: {
     semi: [2, "never"],
-    'no-nonoctal-decimal-escape': 'off'
-  }
+    'no-nonoctal-decimal-escape': 'off',
```
hueitan, line 22: `no-nonoctal-decimal-escape` isn't supported in the current eslint version, please double check if this rule is needed; without this rule also fine.

.eslintrc.js
```diff
@@ -18,6 +19,12 @@ module.exports = {
...
+    "no-only-tests/no-only-tests": "error",
+    "cypress/no-unnecessary-waiting": "warn"
+  },
+  "plugins": [
+    "cypress",
+    "no-only-tests"
+  ],
```
hueitan, line 29: In this file we use single quote `'`, please update all `"` to `'`

hueitan (changes requested):
Overall this is great being the first script in the E2E test, there are some minor changes needed. There is still some error after the fix, try running the lint script and fix it.

.eslintrc.js
```diff
@@ -18,6 +19,12 @@ module.exports = {
   },
   rules: {
     semi: [2, "never"],
-    'no-nonoctal-decimal-escape': 'off'
-  }
+    'no-nonoctal-decimal-escape': 'off',
+    "no-only-tests/no-only-tests": "error",
+    "cypress/no-unnecessary-waiting": "warn"
+  },
+  "plugins": [
```
hueitan, line 26:
```suggestion
  plugins: [
```

.eslintrc.js
hueitan, line 29:
```suggestion
  ]
```

package.json
jpita: I get this error running eslint

`
/Users/josepita/code/wikipedia-preview/cypress/plugins/index.js
  19:1  error  'module' is not defined  no-undef`

but I was checking the file from the kaios repo and it looks exactly the same...

hueitan: I can suggest you do one of the below:

updating it to es6

```diff
- module.exports = ( on, config ) => {
+ export default ( on, config ) => {
```

or adding `module` in as the global var

```js
globals: {
    ...
    module: true
},
```

jpita: the first option is not recognised by Cypress, fails the run.

the second one, where should I add it?

hueitan: `.eslintrc.js`

jpita:
@hueitan can I merge?

medied:
I think I'm missing the full context for this PR, I’m wondering: what will be the relationship between the tests in `/cypress` directory and the ones in `/test` ?

cypress/plugins/index.js
```diff
@@ -0,0 +1,22 @@
...
+// ***********************************************************
+
+// This function is called when a project is opened or re-opened (e.g. due to
+// the project's config changing)
+
+/**
+ * @type {Cypress.PluginConfig}
+ */
+// eslint-disable-next-line no-unused-vars
+module.exports = ( on, config ) => {
```
stephanebisson, line 19: What is this file doing?

cypress/integration/active-spec.js
```diff
@@ -0,0 +1,31 @@
+import { Preview } from '../page-objects/preview-page'
+
+const preview = new Preview()
+
+describe( 'Check the Hover and Click Event', () => {
+
+	beforeEach( 'Open the English Page', () => {
+		cy.navigateToEnglishPage()
```
stephanebisson, line 8: Why is this spec using the English page instead of the test page, which appears to be a copy of the english page?

cypress/integration/layout-spec.js
```diff
@@ -0,0 +1,48 @@
...
+
+	it( 'Check the Wikipedia Preview in Standard with Image Layout', () => {
+		preview.getPreviewSpan().eq( 0 ).click()
+
+		preview.checkPreviewStandard()
+
+		preview.getHeaderClosebtn().click()
+	} )
+
+	it.skip( 'Check the Wikipedia Preview in Offline Layout', () => {
```
stephanebisson, line 20: Why are the next 3 tests skipped?

cypress/integration/layout-spec.js
```diff
@@ -0,0 +1,48 @@
...
+		goOffline()
+
+		preview.getPreviewSpan().eq( 0 ).click()
+
+		preview.checkPreviewOffline()
+
+		preview.getHeaderClosebtn().click()
+
+		goOnline()
```
stephanebisson, line 29: If going back online is important for the other tests in this suite, doing something like `ensureOnline()` in the `beforeEach()` would be more reliable since exceptions in this test would skip the call to `goOnline()` in the last line.

cypress/plugins/index.js
jpita: this is needed for cypress to work I believe, it's the same as in the kaios repo

cypress/integration/active-spec.js
jpita: we created the test page to add other cases (disambiguation, error, offline, etc...), this test uses the english page "just because" I believe, but @shailesh1028 can say more to this

cypress/integration/layout-spec.js
jpita: WIP for the next phases of work since the offline mode doesn't work as expected on cypress

cypress/integration/layout-spec.js
jpita: fixed

stephanebisson (approved):

cypress/integration/active-spec.js
shailesh1028: @jpita You have explained it right. The content of the test file is changed to get the other cases like disambiguation, error, etc. While the English file does not cover the above cases, hence we created the test file.

PR #131, opened 2021-07-22
Page Movement Test

shailesh1028:
The following PR has Test files to check the page movement like reloading the page and moving to the Wikimedia page while reading more .


cypress/integration/expand-spec.js
```diff
@@ -0,0 +1,23 @@
...
+
+	it( 'Check the Wikipedia Preview in Expanded Layout', () => {
+		// Opens the Preview
+		preview.getPreviewSpan().eq( 0 ).click()
+		// Click on Continue Reading Footer element
+		preview.getFooterContiReading().click()
+		// Check the Expanded Mode preview
+		preview.checkPreviewExpanded()
+		// Closes the Preview
+		preview.getHeaderClosebtn().click()
```
jpita, line 19: why do u need this click after the checks?

cypress/page-objects/preview-page.js
```diff
@@ -83,6 +83,13 @@ export class Preview {
 
 	}
 
+	// Custom Command to check the existence of Mobile Preview
+	checkMobilePreview() {
+
```
jpita, line 88: delete this break

jpita (changes requested):

cypress/page-objects/preview-page.js
shailesh1028: Are you referring to the comments?

cypress/integration/expand-spec.js
shailesh1028: It basically checks whether the preview is being opened, by actually closing it. Because the 'checks' only check the visibility of elements only. So we need to make some clicks to assert our case.

cypress/page-objects/preview-page.js
jpita: the empty space after the `{`

PR #132, opened 2021-07-22
Gallery Testing

shailesh1028:
Following Files have been added:
Gallery Page Object: A page object for the gallery of the preview has been created
Gallery-spec: This tests the event of visiting and page movement(next and previous of Images) has been created.
Mob-Gallery-spec: This tests the event of visiting and page movement(next and previous of Images by swiping) in Mobile view, has been created.


cypress/integration/gallery-spec.js
```diff
@@ -0,0 +1,57 @@
...
+		preview.getPreviewSpan().first().click()
+		// Check for the Continue Reading exist
+		preview.getFooterContiReading().then( res => {
+			if ( res.css( 'display' ) !== 'none' ) {
+				preview.getFooterContiReading().click()
+			}
+		} )
+		// Check if the Images exist
+		preview.getBodyGalleryImages().its( 'length' ).then( res => {
+			if ( res > 0 ) {
```
jpita, line 24: if the res == 0 what is this test doing?

cypress/integration/gallery-spec.js
```diff
@@ -0,0 +1,57 @@
...
+			// Open the Gallery and Close it
+				preview.getBodyGalleryImages().first().click()
+				gallery.getGalleryCloseBtn().click()
+			}
```
jpita, line 28: what is this test testing? I see no checks here

cypress/integration/gallery-spec.js
```diff
@@ -0,0 +1,57 @@
...
+				// Open the Gallery
+				preview.getBodyGalleryImages().first().click()
+				// Click on the Next Icon
+				gallery.getGalleryNextBtn().click()
+				// Click on the Previous Icon
+				gallery.getGalleryPrevBtn().click()
+				// Close the Gallery
+				gallery.getGalleryCloseBtn().click()
+			}
+		} )
```
jpita, line 53: what is this test testing? I see no checks here

cypress/integration/mob-gallery-spec.js
```diff
@@ -0,0 +1,89 @@
+import { Preview } from '../page-objects/preview-page'
```
jpita, line 1: change the name for an explicit  name, what is mob?

cypress/integration/mob-gallery-spec.js
```diff
@@ -0,0 +1,89 @@
...
+		preview.getBodyGalleryImages().its( 'length' ).then( res => {
+			if ( res > 1 ) {
+				// Open the Gallery
+				preview.getBodyGalleryImages().eq( 1 ).click()
+				// Check if the Caption Bar exist
+				gallery.getGalleryCaptionBar( i ).its( 'length' ).then( len => {
+					if ( len > 0 ) {
+						// Scroll the Caption Text
+						gallery.getGalleryCaptionText( i ).swipe( 'bottom', 'top' )
+						// gallery.getGalleryCaptionText(i).scrollTo('bottom')
```
jpita, line 82: delete commented code if it's not needed

jpita (changes requested):

cypress/integration/gallery-spec.js
```diff
@@ -0,0 +1,57 @@
...
+	beforeEach( 'Open the English Page', () => {
+		cy.navigateToTestPage()
+	} )
+
+	it( 'Check the Gallery Pages', () => {
+		// Open the Preview
+		preview.getPreviewSpan().first().click()
+		// Check for the Continue Reading exist
+		preview.getFooterContiReading().then( res => {
+			if ( res.css( 'display' ) !== 'none' ) {
```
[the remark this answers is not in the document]
shailesh1028: This is added to make the test case more standard. We are not sure that the images present at the bottom of the content are visible(expanded mode) or not (standard mode). So, I am checking here whether we need to convert it to the expanded mode in order to make the gallery images visible which is done by clicking on the 'continue reading'. So I need to check whether the display of the 'continue reading' is visible or not.

cypress/integration/gallery-spec.js
shailesh1028: This is added to make sure that the images are present in the Gallery. The test is about visiting the gallery mode. Thereby it needs to have at least one image present in the preview content.

cypress/integration/gallery-spec.js
shailesh1028: You asked me to make the test case in such a way that instead of adding checks, we check it by making clicks. In case u say, I will add the checks on the visibility of components.

cypress/integration/mob-gallery-spec.js
shailesh1028: mob refers to mobile. I tried to keep the name short.

cypress/integration/gallery-spec.js
jpita: instead of adding checks in the middle of the test for present elements that are not related to the test and are already checked in other tests.

any test needs to have at least one check at the end of the test

cypress/integration/gallery-spec.js
jpita: since u can add any words to the test.html page, why not add articles with images to be sure they're present?
no need to add code checking that the image is there, this test won't run/test anything if u find an article without images

cypress/integration/gallery-spec.js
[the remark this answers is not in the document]
jpita: since u can add any words to the test.html page, why not add articles with images to be sure the expanded/standard mode is visible?

cypress/integration/mob-gallery-spec.js
jpita: don't save on characters , always make names explicit

cypress/integration/mob-expand-spec.js
```diff
@@ -0,0 +1,38 @@
+import { Preview } from '../page-objects/preview-page'
+
+const preview = new Preview()
+
+describe( 'Check the Expanded Layout in Mobile View', () => {
+
+	beforeEach( 'Open the English Page', () => {
+		cy.navigateToMobileTestPage()
+		cy.viewport( 550, 750 )
```
hueitan, line 9: do you plan to moving this line into the command `navigateToMobileTestPage`? so we don't need to repeat the viewport command in every spec.

shailesh1028: Refactor the code and added the viewport in the command navigateToMobileTestPage

cypress/integration/expand-spec.js
```diff
@@ -0,0 +1,23 @@
+import { Preview } from '../page-objects/preview-page'
+
+const preview = new Preview()
+
+describe( 'Check the Expanded Layout', () => {
+
+	beforeEach( 'Open the English Page', () => {
```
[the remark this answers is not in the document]
shailesh1028: Made the changes accordingly.

cypress/integration/expand-spec.js
```diff
@@ -0,0 +1,23 @@
...
+		cy.navigateToTestPage()
+	} )
+
+	it( 'Check the Wikipedia Preview in Expanded Layout', () => {
+		// Opens the Preview
+		preview.getPreviewSpan().eq( 0 ).click()
+		// Click on Continue Reading Footer element
+		preview.getFooterContiReading().click()
```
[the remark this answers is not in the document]
shailesh1028: Made the changes accordingly.

cypress/integration/gallery-spec.js
```diff
@@ -0,0 +1,57 @@
+import { Preview } from '../page-objects/preview-page'
+import { Gallery } from '../page-objects/gallery-page'
+
+const preview = new Preview()
+const gallery = new Gallery()
+
+describe( 'Check the Gallery Pages ', () => {
+
+	beforeEach( 'Open the English Page', () => {
```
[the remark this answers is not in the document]
shailesh1028: Made the changes accordingly.

cypress/integration/gallery-spec.js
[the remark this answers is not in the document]
shailesh1028: The aim of this condition is:
Firstly, the preview comes up on the available space. Thereby it can be standard or expanded based on the available space. We can't be sure that it will be in standard or expanded mode. This condition just ensures that the preview remains in mode with the gallery images being visible.
Secondly,  While our test is for gallery images, and this is just a condition and has no checks inside it.
lastly, I have a chosen a word that will always fall in the condition and the test never fails.

cypress/integration/gallery-spec.js
shailesh1028: Firstly, the images depend on the content they receive from the API. Suppose there is a change in content in the future then our condition takes care of it. 
Secondly, it just ensures that the proper number of images is there for testing. It is a fallback condition and never disrupts the test.
Lastly, I have chosen a word such that it has images and always falls inside the condition.Hence, it will never fail.
Thereby, I believe that this condition is necessary.

cypress/integration/gallery-spec.js
shailesh1028: Made the changes accordingly. added a check for gallery section

cypress/integration/mob-gallery-spec.js
shailesh1028: Made the changes accordingly.

cypress/integration/mob-gallery-spec.js
shailesh1028: The Swipe command isn't working for the following case. Need your help for the following

PR #133, opened 2021-07-22
T269787: gallery zoom follow up

medied:
Follow up to #116; addresses specific feedback from Sudhanshu: https://phabricator.wikimedia.org/T269787#7214284. The commit list includes all commits from original PR, but the 'Files changed' tab correctly shows only the new additions as the original PR has been already merged with `main`.

### DIFF
--- a/src/gallery/gestures.js
+++ b/src/gallery/gestures.js
@@ -55,6 +55,11 @@ const isImgLandscape = ( image ) => {
 	return image.naturalHeight <= image.naturalWidth
 }
 
+const isImgSmallerThanViewport = ( image ) => {
+	const buffer = 50
+	return image.naturalWidth + buffer < clientWidth
+}
+
 const getFingerAmount = () => {
 	return evCache.length
 }
@@ -75,6 +80,9 @@ const setTransformOrigin = ( image, e ) => {
 	const currentTransformOrigin = getTransformOrigin( e )
 	if ( isImgLandscape( image ) ) {
 		currentTransformOrigin.y = currentTransformOrigin.y - image.naturalHeight
+	} else if ( isImgSmallerThanViewport( image ) ) {
+		currentTransformOrigin.x = image.naturalWidth / 2
+		currentTransformOrigin.y = image.naturalHeight / 2
 	}
 	return `${currentTransformOrigin.x}px ${currentTransformOrigin.y}px`
 }
--- a/src/gallery/gestures.js
+++ b/src/gallery/gestures.js
@@ -68,20 +68,26 @@ const getFingerAmount = () => {
 	return evCache.length
 }
 
-const getTransformOrigin = ( e = null ) => {
+const getTransformOrigin = ( e = null, image ) => {
 	let coordinates = {}
+	const setBestVertical = () => {
+		if ( e.clientY > image.naturalHeight && !isImgLandscape( image ) ) {
+			return image.naturalHeight
+		}
+		return e.clientY
+	}
 	if ( evCache.length === 2 ) {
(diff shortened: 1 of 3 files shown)


hueitan (commented):
Zoom in then moving sometimes has a little jump, I can't tell the exact step, but do you also notice this litle jump?

hueitan:
I can notice this jump, when double tap to zoom in, then double tap again to zoom out.

medied:
I think I can see a tiny jump sometimes, when zooming in and moving, not so much when double-tapping. Would you say this is a blocker? Hasn't come up in the design review

hueitan:
double tap zoom in zoom out ✔️ 

two finger pinch a bit weird to me, it zooms just a little and keep jumping? is this same with you?

tested on iOS safari browser

medied:
Your feedback made me realize that we should reset the `temp.translateX/Y` whenever we clear the zoom, I believe the translate coordinate were getting mixed in between images. Could you try again?

> two finger pinch a bit weird to me, it zooms just a little and keep jumping? is this same with you?

I can share a video of the behavior I'm seeing

stephanebisson:
On my android device, zooming in and out is "slow" in the sense that if I tap with 2 fingers and pinch out, I feel it only zoom about half as much as it should. It doesn't feel like it follows my fingers. Is there a config in the code that translates pinch distance to zoom factor?

medied:
> Is there a config in the code that translates pinch distance to zoom factor?

Hmm not really... there's the zoom factor itself, named `delta` in `zoomMove`, but no method actually translating pinch distance, `delta` remains constant. Notice I adjusted `delta` from 0.1 to 0.01 in this follow up PR, based on design's feedback for a [smoother transition](https://phabricator.wikimedia.org/T269787#7214284). If you change it back to 0.1 for testing, you should notice the difference. 

I'd like to further unpack your feedback though. 

> It doesn't feel like it follows my fingers

Follow your fingers as in the `scale` change itself, as opposed to the `translate3d` coordinates, yes?

Also what browser were you testing with?

hueitan:
> On my android device, zooming in and out is "slow" in the sense that if I tap with 2 fingers and pinch out, I feel it only zoom about half as much as it should. It doesn't feel like it follows my fingers. Is there a config in the code that translates pinch distance to zoom factor?

yes, i'm with @stephanebisson , the 2 fingers pinch zoom is "slow".

When double tap to zoom in, then seems like 2 fingers pinch doesn't work anymore.

hueitan:
I saw one behaviour, not sure this belongs to here or the initial ticket

| Double tap the bottom area | Slide to view the top of the image |
| --- | --- |
| image | image |

You can't view the rest of the top image above the line

medied:
> I saw one behaviour, not sure this belongs to here or the initial ticket
> 
> You can't view the rest of the top image above the line

I'm able to replicate this behavior, I have an idea of something we could do to address it

medied:
@hueitan I updated the zoom scrolling boundaries logic: `getBoundingClientRect` returns the [actual coordinates and numbers of the resized image](https://stackoverflow.com/questions/17763929/webkit-transform-scale-how-to-get-the-coordinates-of-a-visible-area), we can use that to bound the scrolling more appropriately. 

Thank @stephanebisson for committing https://github.com/wikimedia/wikipedia-preview/pull/133/commits/55e474da6513c023f392dc1bff572cdf55d2c2e5. Does that help improve the "slow" pinching behavior you were both seeing?

It's testing decently well from my device.

medied:
We can invoke `getBoundingClientRect` in `zoomStart` only, no need to invoke in every `zoomMove`. Checkout the difference:

image

PR #134, opened 2021-07-30
Support pages in other namespaces

stephanebisson:
https://phabricator.wikimedia.org/T287688

For pages outside the main article namespace, the PCS summary endpoint give a "no-extract" response so the code falls back to the error state. 

This PR tries to improve the situation by using the MediaWiki extract API for those pages. It is also trying to use the "description" field for article returning "no-extract" for whatever reason.

Also added a few unit tests.


src/api.js
```diff
@@ -17,6 +19,16 @@ const requestPagePreview = ( lang, title, isTouch, callback, request = cachedReq
 				type: data.type
 			}
 		}
+		// special case: there is no summary but there is a description
+		if ( data.type === 'no-extract' && data.description ) {
```
hueitan, line 23: Yes, it helps with `no-extract` case and better than nothing.

src/api.js
```diff
@@ -1,14 +1,41 @@
...
+		exsentences: 4,
+		explaintext: 1,
+		exsectionformat: 'plain',
+		piprop: 'thumbnail',
+		pilimit: 1,
+		titles: title
+	}
+	const url = buildMwApiUrl( lang, params )
+	request( url, ( result ) => {
+		const page = result.query.pages[ Object.keys( result.query.pages )[ 0 ] ]
```
hueitan, line 21: We need to filter if there is `missing` key in the response, otherwise the undefined looks super bad

| image | image |
| --- | --- |

hueitan (approved):

hueitan:
I'm trying the summary and this extract query endpoint in article `Food: Fact or Fiction?`

(1) https://en.wikipedia.org/api/rest_v1/page/summary/Food%3A_Fact_or_Fiction%3F?wprov=wppw1
(2) https://en.wikipedia.org/w/api.php?format=json&formatversion=2&origin=*&action=query&prop=extracts%7Cpageimages&exsentences=4&explaintext=1&exsectionformat=plain&piprop=thumbnail&pilimit=1&titles=Food%3A_Fact_or_Fiction%3F

The only lacking of (2) is the thumbnail url, and i can't find it return any kind of thumbnail url like (1) 

And i think this is fine, right?

stephanebisson:
> The only lacking of (2) is the thumbnail url, and i can't find it return any kind of thumbnail url like (1)

The thumbnail is provided by the PageImages extension but it doesn't work very well. It is returning the right image for the Cat article so we know it works in some cases, at least. 

https://en.wikipedia.org/w/api.php?format=json&formatversion=2&origin=*&action=query&prop=extracts%7Cpageimages&exsentences=4&explaintext=1&exsectionformat=plain&piprop=thumbnail&pilimit=10&titles=Cat

src/api.js
```diff
@@ -17,11 +47,27 @@ const requestPagePreview = ( lang, title, isTouch, callback, request = cachedReq
...
+		}
 		return false
 	}, callback )
 
 }
 
+const requestPagePreview = ( lang, title, isTouch, callback, request = cachedRequest ) => {
+	return title.indexOf( ':' ) === -1 ?
+		requestPcsSummary( lang, title, isTouch, callback, request ) :
+		requestMwExtract( lang, title, isTouch, callback, request )
```
medied, line 68: Looks like `isTouch` is not really needed for either `requestMwExtract` or `requestPcsSummary`?

stephanebisson: Right. I removed it.

medied (approved):
Can't think of any other failing cases

jpita:
it's getting the images but no text, is that OK?
using `detectLinks` with https://en.wikipedia.org/wiki/Wikipedia%3AContents%2FGeography_and_places 
image

stephanebisson:
It's not great but it's expected since this page is just a giant table and the text extract extension is unable to find anything.

Note that this page is a very bad fit for Wikipedia Preview.

PR #139, opened 2021-08-04
detectLinks: Leave href alone for right-click options

stephanebisson:
https://phabricator.wikimedia.org/T288034

The `detectLinks` init option replaces the href attribute value with the standard no-op `javascript:;` in touch mode. It makes it impossible to use the standard right click options (like open in new tab) on devices that have both touch support and a pointer.

I removed the href replacement. I don't know why it was there. Please test carefully.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -146,8 +146,6 @@ function init( {
 					node.setAttribute( 'data-wp-title', matches.title )
 					node.setAttribute( 'data-wp-lang', matches.lang )
 					if ( isTouch ) {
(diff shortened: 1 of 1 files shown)


medied (approved):
The bahasa indonesian sample page uses `detectLinks` for "Lombok". I tested that (along with others) both on desktop and directly on device and it's working for me.

hueitan:
It works expected on all my environment (touchable desktop, mouse in desktop, mobile device).

PR #144, opened 2021-08-30
T284601: Expose plugin helper method getPreviewHtml

medied:
https://phabricator.wikimedia.org/T284601

### Objective

This introduces a helper util method to support https://github.com/wikimedia/wikipediapreview-wordpress/pull/30. In order to test this all locally, we have to make use of `npm link` to link the two repos with each other:
  * Follow these steps to set up the `npm link` between the two repos: https://docs.npmjs.com/cli/v7/commands/npm-link
  * then npm run build in wikipedia-preview in branch `T284601-gutenberg-preview`
  * then npm run build in wikipediapreview-wordpress in the branch with the same name
  * then manually copy the production file to assets: `cp node_modules/wikipedia-preview/dist/wikipedia-preview.production.js assets`
  * then you should be able to see the updates reflected in the wikipediapreview-wordpress set up


src/utils.js
```diff
@@ -72,8 +74,16 @@ const version = () => {
 	console.log( `Wikipedia Preview - version ${APP_VERSION} (${GIT_HASH})` )
 }
 
+const getPreviewHtml = ( title, lang ) => {
```
stephanebisson, line 77: Having this inside `utils` is a bit out of place. Utils are low-level. no-dependencies little helpers. They are used by higher level code like api and preview. Now this creates a circular dependency. 

I would recommend creating this function directly in index or in a new file if you can find a good name for it.

src/utils.js
```diff
@@ -72,8 +74,16 @@ const version = () => {
...
+	return new Promise( ( resolve )=> {
```
stephanebisson, line 78: Until we officially remove IE11 support from this product, we can't use promises. You can use a callback like we do in the API layer.

stephanebisson (changes requested):

src/utils.js
medied: Makes sense, let's go with directly in index for now

hueitan: > if you can find a good name for it.

yeah, i think the good name here is very important, as this expose the method to everyone with `wikipediaPreview.getPreviewHtml()`, at least we need write some comment/doc to tell why this is here.

src/index.js
```diff
@@ -17,6 +17,12 @@ const invokeCallback = ( events, name, params ) => {
 	}
 }
 
+const getPreviewHtml = ( title, lang, callback ) => {
+	requestPagePreview( lang, title, isTouch, data => {
```
medied, line 21: Is it worth adding some basic truthy input validation here for `title` and `lang` before executing `requestPagePreview`?

src/utils.js
stephanebisson: It's for internal use only. We can comment on the function itself that it is used by the plugin so we remember not to break it but no public documentation.

src/index.js
stephanebisson: I feel this kind of situation should be handled upstream but not calling this function unless all inputs are available.

src/utils.js
```diff
@@ -72,8 +74,16 @@ const version = () => {
...
+		requestPagePreview( lang, title, isTouch, data => {
```
hueitan: https://github.com/wikimedia/wikipedia-preview/blob/72b6f03bd6c9296de224a5807a13719ce37384b7/src/api.js#L68-L72

the third param should be callback on `requestPagePreview`

_interesting, it works without the problem previously_

medied: Whooa good catch, how was that working before? From the plugin side I was actually providing only three arguments, third one being the function, so I guess JS was smart enough to map the functions and omit the boolean?

PR #146, opened 2021-08-31
Log errors to the console

stephanebisson:
The error state shouldn't happen too often but when it does, it's almost certainly caused by a configuration error. 

With this PR, editors and integrators get a little more information that they can try to understand or send us when they are stuck with unexpected error states.

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -6,6 +6,7 @@
     "gallery-loading-still": "Still loading",
     "gallery-unknown-author": "Author unknown",
     "preview-error-message": "There was an issue while displaying this preview.",
+    "preview-console-error-message": "Preview unavailable for article '$1' (Language: $2)",
     "read-on-wiki": "Read on Wikipedia",
     "read-more": "Read more on Wikipedia",
     "preview-disambiguation-message": "Title <strong>$1</strong> is related to more than one article on Wikipedia.",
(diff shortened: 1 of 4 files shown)


medied (approved):
LGTM 👍

jpita:
[Main_page](https://en.wikipedia.org/wiki/Main_Page) shows the error but the [same page in pt](https://pt.wikipedia.org/wiki/Wikip%C3%A9dia:P%C3%A1gina_principal) still tries to show the preview and doesn't show the error
image

is this expected?

the API calls are also different, for Main_page

https://en.wikipedia.org/api/rest_v1/page/summary/Main_Page?wprov=wppw1

for PT version

https://pt.wikipedia.org/w/api.php?format=json&formatversion=2&origin=*&action=query&prop=extracts%7Cpageimages&exsentences=4&explaintext=1&exsectionformat=plain&piprop=thumbnail&pilimit=1&titles=Wikip%C3%A9dia%3AP%C3%A1gina_principal

and

https://pt.wikipedia.org/api/rest_v1/page/media-list/Wikip%C3%A9dia%3AP%C3%A1gina_principal

what makes it different?

stephanebisson:
It looks like it's working as expected. Pages in other namespaces will work but there's a good chance that the preview is not really helpful.

PR #149, opened 2021-09-27
T289471: Add bdi tags for gallery text

medied:
https://phabricator.wikimedia.org/T289471

Embedding the `<bdi>` within the existing `<div>` (instead of replacing the existing `<div>` with `<bdi>`) because in testing I’m seeing that replacing the `<div>` messes with the the text placement for the author. 

image


hueitan:
in the second screenshot, i think what we are missing is setting `text-align: right` for RTL?

PR #157, opened 2022-06-15
Use fixed position on desktop

stephanebisson:
https://phabricator.wikimedia.org/T306663

On this [site](https://gorhambury.org/), the body has `position: relative` and that's throwing the positioning logic off. 

On this other [site](https://diff.wikimedia.org/2022/06/06/in-search-of-the-least-viewed-article-on-wikipedia/), positioning is strangely but the popup's `bottom` property is often set to a large negative value and I have no idea why it works.

This PR tries to simplify positioning by using only `top`, `left`, `width` and `height` of the popup and the target and using fixed position so it is not relative to any specific element.

Scroll position doesn't appear to be needed so it was removed.

I tested by adding a very large padding to the body of the demo articles so I can scroll around and make sure the popup is positioned correctly next to the target.

Unit tests also updated.

### DIFF
--- a/src/popup.js
+++ b/src/popup.js
@@ -3,23 +3,22 @@ import '../style/popup.less'
 let popup
 
 const computePopupPosition = (
-	targetRect, popupWidth, scrollX, scrollY, innerWidth, innerHeight
+	targetRect,
+	popupWidth, popupHeight,
+	innerWidth, innerHeight
 ) => {
-	let left
-	let right = ''
-	let top = ''
-	let bottom = ''
+	const targetCenterX = targetRect.left + targetRect.width / 2
+	const targetCenterY = targetRect.top + targetRect.height / 2
 
-	left = targetRect.left > ( innerWidth / 2 ) ?
-		( scrollX + targetRect.right - popupWidth ) :
-		( scrollX + targetRect.left )
+	const left = targetCenterX > innerWidth / 2 ?
+		targetRect.left + targetRect.width - popupWidth :
+		targetRect.left
 
-	if ( targetRect.top > ( innerHeight / 2 ) ) {
-		bottom = ( innerHeight - targetRect.top - scrollY )
-	} else {
-		top = ( scrollY + targetRect.bottom )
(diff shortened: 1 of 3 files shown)


medied:
Working well to prevent the original bug of the popup being partially visible sometimes. Only thing I’m seeing is the popup sometimes ends up with some distance from the target word (like in the french article [1]) and sometimes it covers target word (like in the thai article [2])

image

stephanebisson:
> Working well to prevent the original bug of the popup being partially visible sometimes. Only thing I’m seeing is the popup sometimes ends up with some distance from the target word (like in the french article [1]) and sometimes it covers target word (like in the thai article [2])

Good catch! That's very strange. Need to investigate those cases.

hueitan:
> > Working well to prevent the original bug of the popup being partially visible sometimes. Only thing I’m seeing is the popup sometimes ends up with some distance from the target word (like in the french article [1]) and sometimes it covers target word (like in the thai article [2])
> 
> Good catch! That's very strange. Need to investigate those cases.

That's because the content is less than the given space, so it will call `onExpand` and set the max-height, therefore this result; edge case ..

stephanebisson:
> for the position: fixed, the popup doesn't stick with the highlighted text when scroll...

I think that's ok

stephanebisson:
> > > Working well to prevent the original bug of the popup being partially visible sometimes. Only thing I’m seeing is the popup sometimes ends up with some distance from the target word (like in the french article [1]) and sometimes it covers target word (like in the thai article [2])
> > 
> > 
> > Good catch! That's very strange. Need to investigate those cases.
> 
> That's because the content is less than the given space, so it will call `onExpand` and set the max-height, therefore this result; edge case ..

Exactly. Setting the bottom value when the popup is above the target allows it to resize and stay at the right place. It's also true when you click the "read more" button to expand the popup. I think I will bring back the bottom alignment to fix this.

hueitan (approved):
The latest change works and fix the problem!

PR #158, opened 2022-06-16
Guard against incomplete api response

stephanebisson:
https://phabricator.wikimedia.org/T301345

On this [page](https://diff.wikimedia.org/2022/01/29/czech-expert-wikipedians-help-newcomers-through-wikiclub-and-wikiporadna/), invalid links are showing the string "undefined" instead of the standard error message.

In the `requestMwExtract` function, check that the "extract" property exist in the API response.

In the `requestPageMedia` function, make sure the items property exist before trying to call reduce on it.


medied (approved):
Nice

hueitan (approved):

PR #163, opened 2022-06-22
Fix encoded uri issue

medied:
https://phabricator.wikimedia.org/T301345

We learned from this [1] page that we have to decode URIs [2]

[1] https://diff.wikimedia.org/2022/01/29/czech-expert-wikipedians-help-newcomers-through-wikiclub-and-wikiporadna/
[2] https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURI#decoding_a_cyrillic_url

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -1,5 +1,13 @@
 import DOMPurify from 'dompurify'
 
+const recursiveDecodeUri = ( uri ) => {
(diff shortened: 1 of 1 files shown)


hueitan:
Can you add the test case in utils.test? so we know the util can handle encoded link

PR #170, opened 2023-02-15
T329512: Add wordpress class selector

medied:
[T329512](https://phabricator.wikimedia.org/T329512)

Adds `.wp-block` selector so that link styling gets correctly picked up by Wordpress plugin. Context: https://github.com/WordPress/gutenberg/issues/41821. I tested by adding it locally directly to the link CSS file in the plugin.


hueitan (approved):
This works for me 🎉

stephanebisson (changes requested):
wikipedia-preview has no knowledge of WordPress. The fix belongs in the wikipediapreview-wordpress repo.

PR #177, opened 2023-11-09
Modernize build dependencies

medied:
https://phabricator.wikimedia.org/T350698


package.json
```diff
@@ -2,37 +2,43 @@
...
-    "build": "npm run build:dev && npm run build:prod",
-    "dev": "webpack-dev-server --mode development --host=0.0.0.0 --disable-host-check",
-    "storybook": "start-storybook -p 6006",
-    "build-storybook": "build-storybook",
+    "build": "vite build",
+    "dev": "vite",
+    "preview": "vite preview",
+    "storybook": "storybook dev -p 6006",
+    "build-storybook": "storybook build",
     "wba": "npm run build:prod && npx webpack-bundle-analyzer dist/stats.json",
```
medied, line 27: Anyone know where is this `wba` command use? Trying to understand if it's safe to remove. Looks like it was added by Pita a couple of year ago

package.json
```diff
@@ -41,29 +47,26 @@
...
     "jsdom": "^16.4.0",
     "jsdom-global": "^3.0.2",
     "less": "^3.12.2",
     "less-loader": "^5.0.0",
-    "less-plugin-inline-svg": "^2.0.1",
-    "mini-css-extract-plugin": "^0.9.0",
     "mocha": "^7.2.0",
     "nyc": "^15.1.0",
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0",
```
medied, line 59: According to https://github.com/wikimedia/wikipedia-preview/pull/177/commits/f1e0aa2b863d07ae795ce3bc42607b4e8cc7e800 it looks like react was added as a consequence of the storybook update. @hueitan can you confirm this is fine and we should leave as dependency?

src/index.js
```diff
@@ -4,6 +4,7 @@ import { createPopup } from './popup'
 import { createTouchPopup } from './touchPopup'
 import { renderPreview, renderLoading, renderError, renderDisambiguation, renderOffline } from './preview'
 import { getWikipediaAttrFromUrl, buildWikipediaUrl, isTouch, getDir, isOnline, version, getAnalyticsQueryParam } from './utils'
+import '../style/link.css'
```
hueitan, line 7: hmm i'm not sure whether this belongs here, we used to have the linkStyle.js as the external script, and include the css seperately.

medied: That's right, in the webpack config we were setting up two entry points like this:

```
const config = {
  entry: {
    'wikipedia-preview': './src/index.js',
    'default-link-style': './src/linkStyle.js'
  }
...
```

So the idea of including link.css in index.js was to just have one entry point in our new vite config (and therefore only load wikipedia-preview.css for all styling). But if it's better to have multiple entry points we can look into setting up something equivalent in vite. Do you remember if there was a reason we had this css separate?

hueitan: the link css is not part of the wikipedia preview library, it's just a plain link css for the demo website, also use as default link style in wordpress, anyone can replace this css, docs is here
https://github.com/wikimedia/wikipedia-preview#wmf-wp-with-preview

package.json
hueitan: yes i remember this, it's useful to understand how our lib size
> Visualize size of webpack output files with an interactive zoomable treemap.

I found the version in vite, you can now running `npm run vba` to see the result
https://github.com/wikimedia/wikipedia-preview/pull/177/commits/161fcf6592f956812aa49ddbd5506c823f9082f8

demo/articles/arabic.html
```diff
@@ -52,7 +52,7 @@
             </p>
         </div>
         <!-- Scripts -->
-        <script src="../wikipedia-preview.development.js"></script>
```
hueitan, line 55: we are not able to use this in `npm run dev`, this is only available when we run `npm run build`

the solution can be script target="module" to import it to use; then when we deploy to demo site, we need to replace it to normal script (either umd or iife format)

demo/articles/arabic.html
```diff
@@ -52,7 +52,7 @@
...
+        <script src="/dist/wikipedia-preview.umd.cjs"></script>
```
hueitan, line 55: we are not able to use this in `npm run dev`, this is only available when we run `npm run build`

the solution can be script target="module" to import it to use; then when we deploy to demo site, we need to replace it to normal script (either umd or iife format)

package.json
hueitan: removed, we don't write jsx form in the storybook

demo/articles/arabic.html
medied: How about setting `"dev": "vite build && vite",` in the package.json scripts?

src/index.js
hueitan: wait, this may not be a bad idea overall, in many cases we recommend to [include the link css file](https://github.com/wikimedia/wikipedia-preview#wmf-wp-with-preview), if we include it here then we only need to explain how to overwrite the default link style.

_note: the linkStyle.js is just a simple css and doesn't have any bundle size issue._

medied: Okay that makes sense, so that could be part of the documentation update we have in TODO list then

hueitan: https://github.com/wikimedia/wikipedia-preview/pull/177/commits/4e5fd59be3b13e47b343c29b1a06a61d60830e7d

demo/articles/arabic.html
hueitan: that's right, this will work like what we have in webpack version, https://github.com/wikimedia/wikipedia-preview/pull/177/commits/adc9d1c6b3f51cc8d3f55bc9967c1b9f21573c44

src/event.js
```diff
@@ -26,29 +26,29 @@ export const customEvents = popup => {
...
 
 		if ( toElement !== previewElement && !popup.element.contains( toElement ) ) {
-			let timeoutId
+			const timeoutId = addTimeout( popup.hide, 300 )
 			const persistPopup = () => {
 				clearTimeout( timeoutId )
 			}
 
-			timeoutId = addTimeout( popup.hide, 300 )
+			// timeoutId = addTimeout( popup.hide, 300 )
```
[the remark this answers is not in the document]
medied: Updated in line 41 to satisfy linter and forgot to remove this commented out code, thanks

src/index.js
```diff
@@ -4,6 +4,7 @@ import { createPopup } from './popup'
 import { createTouchPopup } from './touchPopup'
 import { renderPreview, renderLoading, renderError, renderDisambiguation, renderOffline } from './preview'
 import { getWikipediaAttrFromUrl, buildWikipediaUrl, isTouch, getDir, isOnline, version, getAnalyticsQueryParam } from './utils'
+import './linkStyle'
```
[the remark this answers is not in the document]
medied: Keeping link.css out of the bundling process seems to be a better solution, see https://github.com/wikimedia/wikipedia-preview/pull/177/commits/e72689a458e23475b44bc4eb1fe0a923c786cdd4

.eslintrc.cjs
```diff
@@ -18,14 +14,19 @@ module.exports = {
...
+    'cypress/no-unnecessary-waiting': 'warn',
+    'es-x/no-rest-spread-properties': 'off',
+    'es-x/no-export-ns-from': 'off',
+    'es-x/no-regexp-named-capture-groups': 'off'
   },
   plugins: [
     'cypress',
-    'no-only-tests'
+    'no-only-tests',
+    'unicorn'
```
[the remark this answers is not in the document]
medied: Hmmm looks like it may not be needed, I think I added earlier to quiet some complain. Will test again and remove if successful

stephanebisson (approved):

stephanebisson:
Please review the readme file for filename or API that would need to be updated.

PR #179, opened 2023-11-21
specific node engines version

hueitan:
the package will fail to build if starting from Node.js version 17. the supported version is added in CI config and specified engines in package.json file.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -18,7 +18,7 @@ jobs:
       - name: Use Node.js
         uses: actions/setup-node@v2
         with:
-          node-version: '18'
+          node-version: '17'
 
       - name: Install and Build
         run: |
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -18,7 +18,7 @@ jobs:
       - name: Use Node.js
         uses: actions/setup-node@v2
         with:
-          node-version: '16'
+          node-version: 'latest'
 
       - name: Install and Build
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -18,7 +18,7 @@ jobs:
       - name: Use Node.js
         uses: actions/setup-node@v2
         with:
-          node-version: '16'
+          node-version: '16.20.2'
```
Nikerabbit, line 21: What benefit does setting a specific version here give?

package.json
```diff
@@ -3,6 +3,9 @@
   "version": "1.8.0",
   "description": "Shows Wikipedia article preview in a popup",
   "main": "dist/wikipedia-preview.production.js",
+  "engines": {
+    "node": "<=16.20.2"
```
Nikerabbit, line 7: Why not <17?

.github/workflows/main.yml
hueitan: no specific version are necessary, i was just testing with the latest version of the node 16; changing to major version 16 for aa clean state.

package.json
```diff
@@ -3,6 +3,9 @@
   "version": "1.8.0",
   "description": "Shows Wikipedia article preview in a popup",
   "main": "dist/wikipedia-preview.production.js",
+  "engines": {
+    "node": "<17"
```
stephanebisson, line 7: Is this PR still relevant since this is also included in #177 ?

hueitan: this is still needed before the #177, as we tell the current version need older node version.

#177 has new node version setting

PR #185, opened 2024-02-20
T293594 Show disambiguation extract in preview instead of custom disambiguation view

hueitan:
Phabricator: https://phabricator.wikimedia.org/T293594


stephanebisson:
Do we need to update the storybook?

https://wikimedia.github.io/wikipedia-preview/T293594/storybook/?path=/story/wikipedia-preview--disambiguation

hueitan:
> Do we need to update the storybook?

That's true, add a new story.

PR #186, opened 2024-02-22
T345056: dark mode

medied:
https://phabricator.wikimedia.org/T345056

Demo: https://wikimedia.github.io/wikipedia-preview/T345056-dark-mode/

This the first step for supporting dark mode in the Wordpress plugin: the dark mode feature will be native to the core library and the plugin requests it as necessary. I've made the spanish demo page an example with the dark view, for simplicity.

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,13 +161,14 @@ If you prefer to style them in a way that makes more sense for your context, sim
 ```
 
 #### CSS custom properties
+
 If you wish to adjust the styling of the light/dark theme, you can override the following CSS custom properties to your liking as shown below, under the appropriate [`prefers-color-scheme`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) query.
 
 ```CSS
 @media (prefers-color-scheme: dark) {
 	.wikipediapreview {
 		--wikipediapreview-primary-background-color: #202122;
-	    --wikipediapreview-secondary-background-color: #202122;
+		--wikipediapreview-secondary-background-color: #202122;
 		--wikipediapreview-primary-color: #eaecf0;
 		--wikipediapreview-filter-setting: invert(1);
 	}
--- a/README.md
+++ b/README.md
@@ -76,6 +76,7 @@ popupContainer | DOM Element | `document.body` | Where to put the popup in the D
 detectLinks | Boolean | `false` | Allow Wikipedia hyperlinks to have the popup
 events | Object | `{}` | Custom event handlers: `{ onShow: <fn>, onWikiRead: <fn> }`
 debug | Boolean | `false` | Shows the debug message when `init()` is called
+prefersColorScheme | string | `'detect'` | Sets theme color. Allowed values are 'light', 'dark' and 'detect'. Setting it to 'light' or 'dark' will dictate theme color regardless of [`prefers-color-scheme`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme); setting to 'detect' will render preview according to [`prefers-color-scheme`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme).
 
 Example (custom selector)
 ```html
--- a/README.md
+++ b/README.md
@@ -160,6 +160,20 @@ If you prefer to style them in a way that makes more sense for your context, sim
 }
 ```
 
+#### CSS custom properties
+If you wish to adjust the styling of the light/dark theme, you can override the following CSS custom properties to your liking as shown below, under the appropriate [`prefers-color-scheme`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) query.
+
+```CSS
+@media (prefers-color-scheme: dark) {
+	.wikipediapreview {
+		--wikipediapreview-primary-background-color: #202122;
+	    --wikipediapreview-secondary-background-color: #202122;
+		--wikipediapreview-primary-color: #eaecf0;
+		--wikipediapreview-filter-setting: invert(1);
+	}
+}
+```
+
 ## Acknowledgements/Contributors
 
 This is heavily inspired by [jquery.wikilookup](https://github.com/mooeypoo/jquery.wikilookup) and [Page Previews](https://www.mediawiki.org/wiki/Page_Previews).
--- a/src/preview.js
+++ b/src/preview.js
@@ -54,7 +54,14 @@ const renderPreview = ( lang, data, isTouch, prefersColorScheme ) => {
 			</div>
 		`.trim()
 
-	return render( lang, isTouch, data.dir, getPreviewHeader( lang, imageUrl ), bodyContent, prefersColorScheme )
+	return render(
+		lang,
+		isTouch,
+		data.dir,
+		getPreviewHeader( lang, imageUrl ),
+		bodyContent,
+		prefersColorScheme
+	)
 }
 
 const renderLoading = ( isTouch, lang, dir, prefersColorScheme ) => {
--- a/src/preview.js
+++ b/src/preview.js
@@ -26,16 +26,20 @@ const getPreviewBody = ( type, message, cta ) => {
 `.trim()
 }
 
-const render = ( lang, isTouch, dir, headerContent, bodyContent ) => {
+const getReadOnWikiCta = ( lang, title, isTouch ) => {
+	return `<a href="${ buildWikipediaUrl( lang, title, isTouch ) }" target="_blank" class="wikipediapreview-cta-readonwiki">${ msg( lang, 'read-on-wiki' ) }</a>`
+}
+
+const render = ( lang, isTouch, dir, headerContent, bodyContent, prefersColorScheme ) => {
 	return `
-		<div class="wikipediapreview ${ isTouch ? 'mobile' : '' }" lang="${ lang }" dir="${ dir }">
+		<div class="wikipediapreview ${ isTouch ? 'mobile' : '' } ${ prefersColorScheme === 'light' ? '' : 'dark' }" lang="${ lang }" dir="${ dir }">
 			${ headerContent }
 			${ bodyContent }
 		</div>
 	`.trim()
 }
 
-const renderPreview = ( lang, data, isTouch ) => {
+const renderPreview = ( lang, data, isTouch, prefersColorScheme ) => {
 	const imageUrl = data.imgUrl,
 		bodyContent = `
 			<div class="wikipediapreview-body">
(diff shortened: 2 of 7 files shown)


style/preview.less
```diff
@@ -352,3 +355,23 @@
 	flex-direction: column;
 	justify-content: center;
 }
+
+@media (prefers-color-scheme: dark) {
```
medied, line 359: Please test by setting your device or browser to dark mode, do you see a dark preview when opening the english demo page for example?

style/preview.less
```diff
@@ -344,6 +330,23 @@
 			}
 		}
 	}
+
+	&.dark {
```
medied, line 334: I wondering if there's a way to not repeat the dark CSS here and in the media query. Something like this https://stackoverflow.com/a/21819097 would make sense but not sure if possible with native CSS

stephanebisson: You're in luck! This is a less file and you can write [mixins](https://lesscss.org/features/#mixins-feature)

style/preview.less
hueitan: what if ppl set to `preferColorScheme: light` when device settings to dark mode? it seems this media will overwrite the light scheme. (no .light css was given in css)

style/preview.less
```diff
@@ -83,20 +83,6 @@
...
-			bottom: 35px;
-			left: 0;
-			right: 200px;
-			top: 235px;
-			background: -moz-linear-gradient( top, rgba( 255, 255, 255, 0 ) 0%, #fff 100% );
-			background: -webkit-linear-gradient( top, rgba( 255, 255, 255, 0 ) 0%, #fff 100% );
-			background: linear-gradient( to bottom, rgba( 255, 255, 255, 0 ) 0%, #fff 100% );
-			filter: progid:DXImageTransform.Microsoft.gradient( startColorstr='#00ffffff', endColorstr='#ffffff', GradientType=0 );
-		}
-
```
hueitan, line 99: just want to double confirm that we also remove the gradient for light mode?

medied: Yes that's right, according to phab ticket specs

style/preview.less
medied: This is what we need but for some reason it's not working for me yet, I'm not seeing CSS dark styles being applied at all

medied: Looking into whether it's something to do with our Vite set up, more investigation needed

style/preview.less
medied: That's a good question, the test I have found so far to answer it is using the scheme option with the rendering tab: https://developer.chrome.com/docs/devtools/rendering/emulate-css#emulate_css_media_feature_prefers-color-scheme 

Currently the media query overwrites the light scheme but we can work around that, though your question does open up another one regarding precedence: if we wish to support both the media query and an explicit/direct way to configure, which one should have more priority?

style/preview.less
medied: The mixin is working inside the media query but not inside the .dark class selector for some reason, trying to sort that out (haven't pushed yet just fyi)

style/preview.less
hueitan: the usage of prefersColorScheme during init() is like a "theme" to fit site owner theme, so I think prefersColorScheme has more priority than the system dark mode setting.

we are here mix the "theme" and dark mode, would be nice to use css class name like "theme-dark" to differentiate them, although we share the same styling.

medied: Updated as discussed, please test carefully

stephanebisson:
There are several "line too long" warnings that should be addressed here.

src/stories/preview.stories.js
```diff
@@ -33,6 +33,11 @@ export default {
 		imgUrl: {
 			name: 'Thumbnail URL',
 			control: 'text'
+		},
+		prefersColorScheme: {
+			name: 'Color Scheme',
+			control: 'inline-radio',
+			options: [ 'light', 'dark' ]
```
hueitan, line 40: no detect?

medied: Good catch

stephanebisson (approved):

style/preview.less
```diff
@@ -352,3 +333,63 @@
...
+
+	.wikipediapreview-footer-loading {
+		background-color: #eaecf0;
+	}
+
+	.wikipediapreview-header-wordmark,
+	.wikipediapreview-header-closebtn {
+		filter: unset;
+	}
+}
```
Nikerabbit, line 395: Some of the duplication with less mixins could be avoided with CSS custom properties (are they okay to use in this project?). Example:
```css
/* Define default light theme custom properties */
:root {
  --background-color: #ffffff;
  --text-color: #000000;
}

/* Override custom properties for dark theme based on media query preference */
@media (prefers-color-scheme: dark) {
  :root {
    --background-color: #000000;
    --text-color: #ffffff;
  }
}

/* Explicit class for light theme */
.light-theme {
  --background-color: #ffffff;
  --text-color: #000000;
}

/* Explicit class for dark theme */
.dark-theme {
  --background-color: #000000;
  --text-color: #ffffff;
}

/* Use custom properties */
body {
  background-color: var(--background-color);
  color: var(--text-color);
}
```

medied: I don't see why it wouldn't be okay to use them, I refactored to use CSS custom properties and indeed it seems a bit cleaner (this way seems to be more of a standard - https://web.dev/articles/prefers-color-scheme#stylesheet_architecture), thanks

stephanebisson: `:root` represent the whole document, probably the `<html>` or `<body>` tag. It's important to remember that this is a small component hosted within any website. If we want to use an approach like that, everything has to be prefixed so that even if it bleeds into the host, it has no side effects on it.

stephanebisson: [LESS variables](https://lesscss.org/features/#variables-feature) would probably give you a similar coding style but with actual isolation.

medied: Oh that's a good point, I can update using LESS variables and maybe prefix it anyway as something like `--wikipediapreview-primary-background-color`

README.md
```diff
@@ -76,6 +76,7 @@ popupContainer | DOM Element | `document.body` | Where to put the popup in the D
 detectLinks | Boolean | `false` | Allow Wikipedia hyperlinks to have the popup
 events | Object | `{}` | Custom event handlers: `{ onShow: <fn>, onWikiRead: <fn> }`
 debug | Boolean | `false` | Shows the debug message when `init()` is called
+prefersColorScheme | string | `'detect'` | Sets theme color. Allowed values are 'light', 'dark' and 'detect'. Setting it to 'light' or 'dark' will dictate theme color regardless of [`prefers-color-scheme`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme); setting to 'detect' will render preview according to [`prefers-color-scheme`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme).
```
hueitan, line 79: do we confirm already the default is `detect` or `null`?

style/preview.less
hueitan: is variable (https://lesscss.org/features/#variables-feature) helpful here?

README.md
medied: I set it as `detect` [here](https://github.com/wikimedia/wikipedia-preview/pull/186/files/21f0c48eb229e3117d72848eda22a704dffc2343#diff-bfe9874d239014961b1ae4e89875a6155667db834a410aaaa2ebe3cf89820556R69)

style/preview.less
medied: Testing out it out [like so](https://github.com/wikimedia/wikipedia-preview/commit/b6c0b1ce50b25e5a736a30ff3a591e0ca4dbf26f) is not working properly, not overriding/updating variables as expected. GPT said "no, you cannot update LESS variables inside a media query. LESS variables are compiled into static CSS values during the compilation process, and they cannot be changed dynamically at runtime, including within media queries". Does that sound like hallucination to you? 

In any case, we can also use well prefixed CSS custom vars or LESS mixins, we have options

medied: Leaving it [with prefixes](https://github.com/wikimedia/wikipedia-preview/pull/186/commits/06edf0ff68438978beb596dd7edafbfb01f113ef) for now, let's make a decision next week

src/preview.js
```diff
@@ -26,16 +26,21 @@ const getPreviewBody = ( type, message, cta ) => {
...
-const render = ( lang, isTouch, dir, headerContent, bodyContent ) => {
+const getReadOnWikiCta = ( lang, title, isTouch ) => {
+	return `<a href="${ buildWikipediaUrl( lang, title, isTouch ) }" target="_blank" class="wikipediapreview-cta-readonwiki">${ msg( lang, 'read-on-wiki' ) }</a>`
+}
+
+const render = ( lang, isTouch, dir, headerContent, bodyContent, prefersColorScheme ) => {
+	const colorScheme = prefersColorScheme === 'detect' ? '' : `wikipediapreview-${ prefersColorScheme }-theme`
 	return `
-		<div class="wikipediapreview ${ isTouch ? 'mobile' : '' }" lang="${ lang }" dir="${ dir }">
+		<div class="wikipediapreview ${ isTouch ? 'mobile' : '' } ${ colorScheme }" lang="${ lang }" dir="${ dir }">
```
hueitan, line 36: ${colorScheme} not being used now.

style/preview.less
hueitan: this looks good now i think, but can we still use mixin() here so we don't repeat the --var settings?

Nikerabbit: Do we need to repeat the styles for the "light" variants? Aren't those the defaults?

hueitan: can we also write in README to explain how the site owner custom their dark mode styling?

custom code like this, and explain what's the var there
```
@media (prefers-color-scheme: dark) {
	.wikipediapreview {
		--wikipediapreview-primary-background-color: #202122;
	        --wikipediapreview-secondary-background-color: #202122;
		--wikipediapreview-primary-color: #eaecf0;
		--wikipediapreview-filter-setting: invert(1);
	}
}
```

medied: Great feedback all, thanks

README.md
```diff
@@ -159,6 +160,20 @@ If you prefer to style them in a way that makes more sense for your context, sim
...
 ```
 
+#### CSS custom properties
+If you wish to adjust the styling of the light/dark theme, you can override the following CSS custom properties to your liking as shown below, under the appropriate [`prefers-color-scheme`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) query.
+
+```CSS
+@media (prefers-color-scheme: dark) {
+	.wikipediapreview {
+		--wikipediapreview-primary-background-color: #202122;
+	    --wikipediapreview-secondary-background-color: #202122;
```
hueitan, line 170: indent

README.md
```diff
@@ -159,6 +160,20 @@ If you prefer to style them in a way that makes more sense for your context, sim
 }
 ```
 
+#### CSS custom properties
+If you wish to adjust the styling of the light/dark theme, you can override the following CSS custom properties to your liking as shown below, under the appropriate [`prefers-color-scheme`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) query.
```
hueitan, line 164: line break between ### and text?

hueitan (approved):

PR #189, opened 2024-03-20
T345056: Fix closure issue

medied:
https://phabricator.wikimedia.org/T345056

Another followup: when we update the color scheme on the fly (because of the wordpress plugin dark mode toggle, for example) we need the latest value to be propagated to `showPopup` in order for preview to render with current scheme after re-initializing


hueitan (approved):
works!

PR #190, opened 2024-03-25
Introduce controller to handle most logic

stephanebisson:
Refactor

- Break up big index.js file
- Introduce controller.js with explicit state and status
- Move various functions to utils
- Simplify previews and links registration

Open issues

- There is logic in events.js that relies on state put on the popup (loading, title, lang, etc), which I was trying to get rid of. How can we refactor events.js to put decisions based on state in controller.js?

### DIFF
--- a/src/controller.js
+++ b/src/controller.js
@@ -1,0 +1,177 @@
+import { requestPagePreview } from './api'
+import { customEvents } from './event'
+import { createPopup } from './popup'
+import { createTouchPopup } from './touchPopup'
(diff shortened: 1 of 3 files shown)


src/controller.js
```diff
@@ -0,0 +1,177 @@
...
+		)
+
+		requestPagePreview( lang, title, ( data ) => {
+			// Compare request id in state and closure
+			// to ensure this callback is not outdated
+			if ( requestId !== state.requestId ) {
+				return
+			}
+			if ( state.status === Status.LOADING ) {
+				popup.loading = false
```
hueitan, line 90: this should be set to state.status

PR #191, opened 2024-03-28
Reef

stephanebisson:
Trying out what the codebase could look like with reefjs

Demo: https://wikimedia.github.io/wikipedia-preview/reef

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,6 +10,7 @@
       "license": "MIT",
       "dependencies": {
         "dompurify": "^3.0.6",
+        "reefjs": "^13.0.2",
         "storybook": "^7.5.3"
       },
       "devDependencies": {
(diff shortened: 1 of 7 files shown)


src/store.js
```diff
@@ -0,0 +1,33 @@
...
+		state.data = null
+		requestPagePreview( lang, title, wpStore.receiveContent )
+	},
+
+	receiveContent( state, data ) {
+		state.data = data
+	},
+
+	close( state ) {
+		state.targetId = null
```
hueitan, line 29: when i click on wordmark, i am not sure why close is called here. 🤔 
Screenshot 2024-04-03 at 13 50 33

medied: Hmm I can't replicate that behavior, for me it's closing correctly when clicking the X and not the wodmark

hueitan: Thanks @medied for checking, i can't test it now, I will wait until this branch ready then test this again.

hueitan:
This PR will not be continued and merged at this time, as we plan to implement a more modern solution in the future. Closing this PR for now, but the work done here may inform the future implementation.

PR #192, opened 2024-04-03
T360002 remove font weight setting for link

hueitan:
Phabricator: https://phabricator.wikimedia.org/T360002

this is not being used, anyone can overwrite the styling of the link (including wordpress style)

### DIFF
--- a/demo/css/link.css
+++ b/demo/css/link.css
@@ -1,6 +1,5 @@
 .wmf-wp-with-preview {
   position: relative;
-  color: inherit;
   cursor: pointer;
 }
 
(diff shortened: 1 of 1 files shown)


demo/css/link.css
```diff
@@ -1,6 +1,5 @@
 .wmf-wp-with-preview {
   position: relative;
-  font-weight: 400;
   color: inherit;
```
stephanebisson, line 3: Can you also remove this line since it doesn't do anything

hueitan: DONE.

PR #196, opened 2024-04-25
Place dark theme class names higher up the DOM

medied:
As part of the Wikipedia Preview refactor (https://phabricator.wikimedia.org/T363468), this refactors the dark theme class names placement from `.wikipediapreview` to top level `.wp-popup-container`

### DIFF
--- a/src/components/preview.js
+++ b/src/components/preview.js
@@ -33,12 +33,10 @@ const getReadOnWikiCta = ( lang, title, isTouch ) => {
 	return `<a href="${ buildWikipediaUrl( lang, title, isTouch ) }" target="_blank" class="wikipediapreview-cta-readonwiki">${ msg( lang, 'read-on-wiki' ) }</a>`
 }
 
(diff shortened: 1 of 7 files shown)


hueitan (changes requested):
This change is great, storybook does use this state for preview, can you look into it?

Nikerabbit:
Please add a message why or what this fixes (see also https://www.mediawiki.org/wiki/Gerrit/Commit_message_guidelines)

Nit pick: I guess you mean tree, not a stack? Or could just say higher up in the DOM.

medied:
> This change is great, storybook does use this state for preview, can you look into it?

You're right, it doens't make sense to keep the color scheme test under the preview storybook now as the classes get set in the container. So I've removed them from the preview storybook and added it under the target storybook, what do you think?

medied:
> Please add a message why or what this fixes (see also https://www.mediawiki.org/wiki/Gerrit/Commit_message_guidelines)
> 
> Nit pick: I guess you mean tree, not a stack? Or could just say higher up in the DOM.

Ah yes: higher up the DOM! PR description updated

PR #199, opened 2024-04-29
T363481: Use text-decoration for link style

medied:
Phab ticket: https://phabricator.wikimedia.org/T363481
Demo: https://wikimedia.github.io/wikipedia-preview/T363481-underline/articles/english.html

Wikipedia Preview links styling is currently implemented using a pseudo element (`::after`). The following issues have been found to be likely due to this implementation:

* When the target is on more than one line, the underline disappears
* The gap between the word(s) and the underline is not part of the element so moving the mouse there will cause the popup to disappear

This PR uses `text-decoration` instead to fix issues above, sample below:

image

### DIFF
--- a/demo/css/link.css
+++ b/demo/css/link.css
@@ -1,14 +1,11 @@
 .wmf-wp-with-preview {
   position: relative;
   cursor: pointer;
-}
-
-.wmf-wp-with-preview::after {
-  content: '';
-  position: absolute;
-  left: 0;
-  bottom: -4px;
-  width: 100%;
-  opacity: 0.5;
-  border-bottom: 2px dashed;
+  text-decoration: dashed underline 2px;
+  text-decoration-color: rgba(0, 0, 0, 0.5);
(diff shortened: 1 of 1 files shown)


hueitan (approved):
NIce, this works! 

I wondering myself why changing it to use `::after` in my last commit
https://github.com/wikimedia/wikipedia-preview/pull/174/files

demo/css/link.css
```diff
@@ -1,14 +1,11 @@
...
-.wmf-wp-with-preview::after {
-  content: '';
-  position: absolute;
-  left: 0;
-  bottom: -4px;
-  width: 100%;
-  opacity: 0.5;
-  border-bottom: 2px dashed;
+  text-decoration: dashed underline 2px;
+  text-decoration-color: rgba(0, 0, 0, 0.5);
```
stephanebisson, line 5: Setting a color here doesn't look good in dark mode.

PR #201, opened 2024-04-30
Remove close icon on desktop

stephanebisson:
https://phabricator.wikimedia.org/T363802

Close icon (x) is redundant on desktop since hovering out of the preview closes it. Therefore removing it.


medied (approved):
Looks much better without close icon, thanks for updating tests too

PR #204, opened 2024-05-02
Extract section summary

stephanebisson:
https://phabricator.wikimedia.org/T360286

Storybook: https://wikimedia.github.io/wikipedia-preview/section-api/storybook/?path=/story/sections--sections
Demo: https://wikimedia.github.io/wikipedia-preview/section-api/articles/french.html (first preview: "resolutions" is linking to the "Histoire" section of the article)


hueitan (commented):
it would be nice if we can add the section info within the debug mode
image

hueitan (commented):
we also need to make sure the detectLinks work for section

src/api.js
```diff
@@ -65,13 +69,138 @@ const requestPcsSummary = ( lang, title, callback, request = cachedRequest ) =>
...
+		for ( const section of info.sections ) {
+			if ( section.id === sectionId ) {
+				callback( {
+					title: title + '#' + sectionId,
+					extractHtml: section.extractHtml,
+					imgUrl: section.imgUrl,
+					dir: info.dir,
+					type: 'standard'
+				} )
+			}
```
hueitan, line 179: question, the callback doesn't break here?

medied: Yes good catch, it should return/break after the matching section callback, otherwise we always end up with `No preview available for this section` in the Wordpress plugin selector

src/api.js
```diff
@@ -65,13 +69,143 @@ const requestPcsSummary = ( lang, title, callback, request = cachedRequest ) =>
...
+					imgUrl: section.imgUrl,
+					dir: info.dir,
+					type: 'standard'
+				} )
+				return
+			}
+		}
+		callback( {
+			title,
+			extractHtml: '<p>No preview available for this section</p>',
```
hueitan, line 188: do we need to i18n this line?

hueitan (approved):
LGTM works very well :tada:

stephanebisson:
> we also need to make sure the section summary works for the detectLinks

I had missed this comment. I made changes to the regexp and test to support this.

PR #205, opened 2024-05-06
WIP - replace compute position using floating-ui

hueitan:


hueitan:
one problem using this lib, the popup will show on the center when having multiple lines.

Screenshot 2024-05-07 at 13 58 58

hueitan:
one more problem in storybook, example for iframe
image

src/preview.js
```diff
@@ -30,17 +30,20 @@ const getReadOnWikiCta = ( lang, title, isTouch ) => {
 	return `<a href="${ buildWikipediaUrl( lang, title, isTouch ) }" target="_blank" class="wikipediapreview-cta-readonwiki">${ msg( lang, 'read-on-wiki' ) }</a>`
 }
 
-const render = ( lang, isTouch, dir, headerContent, bodyContent, prefersColorScheme ) => {
+const render = (
+	lang, isTouch, dir, headerContent, bodyContent, prefersColorScheme, withArrow = true
```
stephanebisson, line 34: Is there anyway we could make the arrow part of the popup component instead of the preview?

package.json
```diff
@@ -64,6 +64,7 @@
     "vitest": "^0.34.6"
   },
   "dependencies": {
+    "@floating-ui/core": "^1.6.1",
```
stephanebisson, line 67: Can you confirm that this is exactly the version that is approved for use in codex? 

Also, let's remove the ^ to make sure we always pull the same version.

src/popup.js
```diff
@@ -75,27 +25,51 @@ const createPopup = ( container, win = window ) => {
...
+
+		computePosition( nextTo, popup, {
+			middleware: [
+				inline( { x: mouseX, y: mouseY } ),
+				shift(),
+				autoPlacement(),
+				offset( 10 ),
+				arrow( { element: arrowEl } )
+			]
+		} ).then( ( { x, y, middlewareData, placement } ) => {
```
stephanebisson, line 40: I this that this is async with a promise.
1. We've not used promises in this codebase at all for compat reasons. This is probably fine now but we'll have to re-evaluate our supported browsers list and redo some compat testing. This is just a note.
2. Is there a sync interface we could use instead to make it easier to integrate with the reef branch?

stephanebisson (changes requested):

package.json
hueitan: make sense

codex is using `@floating-ui/vue` 1.0.6, latest version as well, 

we actually need `@floating-ui/dom` here, should we submit another security/performance ticket?

src/popup.js
hueitan: good point, the source code uses a lot async await code

package.json
stephanebisson: It's using the same core component so I think we are ok

src/preview.js
hueitan: Done, please check it again.

style/preview.less
```diff
@@ -376,3 +380,7 @@
 	--wikipediapreview-primary-color: #202122;
 	--wikipediapreview-filter-setting: unset;
 }
+
+.wikipediapreview + .wp-popup-arrow {
+	background-color: var( --wikipediapreview-primary-background-color );
+}
```
stephanebisson, line 386: Missing a new line at the end of this file

stephanebisson (commented):
Works well. 

One big difference is that it shows the popup right or left of the target whereas on main is shows mostly above of below. I think it looks better but it's something we should bring up with Sudhanshu before merging.

stephanebisson (approved):

PR #210, opened 2024-05-22
Simplify reader interactions

medied:
https://phabricator.wikimedia.org/T360752

Demo: https://wikimedia.github.io/wikipedia-preview/T360752-reader-interactions/

* Adds a scroll cue (fade) without adding event listeners
* Removes a bunch of code related to onExpand and simplifies the reader interaction
* Updates Storybook and tests accordingly

---

image

---

- [x] leave gallery row
- [x] [fix cypress tests](https://github.com/wikimedia/wikipedia-preview/actions/runs/9193275427/job/25283909266?pr=210)
- [x] update storybook
- [x] update README screenshots


hueitan (changes requested):
The gallery row are being called twice after the 2nd time of hovering the highlighted text, we used to call it when onExpand, but now this is changed.
image

style/preview.less
```diff
@@ -235,66 +235,60 @@
...
+			height: 11px;
+			background-image: url( '../images/link-external-ltr.svg' );
+			background-repeat: no-repeat;
 		}
 	}
 
 	&.mobile {
 		width: 100%;
-	}
+		height: 55vh;
```
hueitan, line 288: this should be max-height, the shorter article content can be less than 55vh

hueitan:
| first time when you hover a disambiguation text <br/> (cue is available) | second time it works fine but showing console error | 
| --- | --- |
| image | image |

medied:
Good catch Huei, I've updated to address your feedback. Note the code flow related to the gallery will likely be further updated when we work on https://phabricator.wikimedia.org/T360753

style/preview.less
```diff
@@ -83,13 +83,13 @@
...
-		overflow: hidden;
+		overflow: scroll;
 
 		p {
 			margin: 0;
-			line-height: 1.6;
+			line-height: 1.7;
 			font-size: 18px;
-			padding: 10px 20px;
+			padding: 5px 20px;
```
hueitan, line 92: from the design sharing in the phab i can see the padding should remain as padding: 10px 20px; can you confirm that again?

medied: Updated 👍

src/preview.js
```diff
@@ -27,7 +27,23 @@ const getPreviewBody = ( type, message, cta ) => {
 }
 
 const getReadOnWikiCta = ( lang, title, isTouch ) => {
-	return `<a href="${ buildWikipediaUrl( lang, title, isTouch ) }" target="_blank" class="wikipediapreview-cta-readonwiki">${ msg( lang, 'read-on-wiki' ) }</a>`
+	return `<a href="${ buildWikipediaUrl( lang, title, isTouch ) }" target="_blank" class="wikipediapreview-footer-link-cta">${ msg( lang, 'read-on-wiki' ) }</a>`
+}
+
+const getLinkIconSvg = ( dir = 'ltr' ) => {
```
medied, line 33: For security reasons the rules you can set to `::visited` are very limited, only color related changes basically, so we can't add the svg as `content` for `::after` if we want to achieve the color change

medied: Great thanks, `getLinkIconSvg` added to utils and dark mode question brought up to design already

src/stories/preview.stories.js
```diff
@@ -53,33 +52,11 @@ export default {
 }
 
 export const StandardWithImage = ( { lang, title, extractHtml, dir, pageUrl, imgUrl, touch, prefersColorScheme } ) => {
-	return renderPreview( lang, { title, extractHtml, dir, pageUrl, imgUrl }, touch, prefersColorScheme )
+	return `<div style="position: absolute;">${ renderPreview( lang, { title, extractHtml, dir, pageUrl, imgUrl }, touch, prefersColorScheme ) }</div>`
```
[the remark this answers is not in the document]
medied: This is to replicate the concept of a decorator but manually and directly with the string output that renderPreview provides, as decorators are for actual components per se. The scroll-cue is positioned absolute to the bottom, which in the Storybook overspills in the rendering space, so I added the wrapper div to contain the cue and have it displayed in the right place

https://storybook.js.org/docs/writing-stories/decorators

PR #211, opened 2024-05-22
Validate and default popupContainer

stephanebisson:
https://phabricator.wikimedia.org/T361703

Validate `popupContainer` and default it to `document.body` before trying to use it.


hueitan (approved):
LGTM

PR #213, opened 2024-05-31
Hide scrollbars

stephanebisson:
https://phabricator.wikimedia.org/T364881

* Favor `overflow: auto` over `overflow: scroll` since the former will only show scrollbars if they are needed
* Try to remove scrollbars with webkit and firefox specific attributes

### DIFF
--- a/style/gallery.less
+++ b/style/gallery.less
@@ -10,7 +10,7 @@
 
 		&-row {
 			display: flex;
-			overflow: scroll;
+			overflow: auto;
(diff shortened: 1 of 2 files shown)


hueitan (approved):
Nice with the `scrollbar-width`

medied (approved):
Some of these might conflict with https://github.com/wikimedia/wikipedia-preview/pull/210, I'll merge and rebase the other PR

PR #215, opened 2024-06-05
Support computers with touch screen

stephanebisson:
Phab: https://phabricator.wikimedia.org/T257574
Demo: https://wikimedia.github.io/wikipedia-preview/desktop-touch/

* Popup position based on width
* mouseenter event for both pointer and touch


hueitan (approved):
it works for me 🎉

medied (approved):

PR #218, opened 2024-06-18
T360753: Gallery row update

medied:
https://phabricator.wikimedia.org/T360753

Demo: https://wikimedia.github.io/wikipedia-preview/T360753-gallery-images/articles/english.html

* Updates preview layout 
* Loading state will be maintained until both preview and media data are available
* Storybook and tests updated accordingly

---

image

### DIFF
--- a/style/preview.less
+++ b/style/preview.less
@@ -72,8 +72,8 @@
 			background-repeat: no-repeat;
 			background-position: center;
 			cursor: pointer;
-			height: 100%;
-			text-align: center;
+			align-self: center;
+			height: 50px;
 			width: 50px;
 			filter: var( --wikipediapreview-filter-setting );
 		}
--- a/style/preview.less
+++ b/style/preview.less
@@ -93,8 +93,8 @@
 
 		p {
 			margin: 0;
-			line-height: 1.5;
-			font-size: 18px;
+			line-height: 1.6;
+			font-size: 16px;
 			padding: 0 20px;
 			color: var( --wikipediapreview-primary-color );
 		}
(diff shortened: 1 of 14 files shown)


hueitan:
the gallery height looks a little bit bigger than usual?
image

medied:
> the gallery height looks a little bit bigger than usual?

I know right? And yet it's still about 10px smaller than the design in Figma. I plan to sync with Sudhanshu to go over the pixel details

style/preview.less
```diff
@@ -20,18 +20,18 @@
 
 	&-header {
 		display: flex;
-		height: 78px;
+		height: 52px;
 		line-height: 62px;
-		margin-top: 12px;
-		margin-bottom: 12px;
```
hueitan, line 26: | the header with thumbnail top margin/padding doesn't look good on | header without thumbnail looks fine | header get cutoff in smaller screen (i think it doesn't happen before) |
| ---| --- | --- |
| image| image |image |

hueitan (commented):
| Before | After |
| --- | --- |
| image | image |

\
seems like loading bar are getting smaller, why is that?

hueitan:
One more thing, should the max-height be the same for case with and without thumbnail, and more content scrollable

currently it doesn't look good to me
| with thumbnail | without thumbnail |
| --- | --- |
| image | image |

medied:
Thanks @hueitan, I synced with Sudhanshu today and went over all your feedback, please check the latest style update

hueitan (approved):
merge it now

